PR #1, opened 2018-10-24
less fragile environment.yml

rsignell-usgs:
Here the only channel specified at the top of the `environment.yml` is `conda-forge`, and if we need a package from another channel we specify that as part of the package requirement.   This should help the conda solver do a better job at pulling from `conda-forge` if possible.  

This should also pull the bokeh 1.0 release from conda-forge, which  just arrived 20 min ago!

### DIFF
--- a/deployments/dev.pangeo.io/image/binder/environment.yml
+++ b/deployments/dev.pangeo.io/image/binder/environment.yml
@@ -1,28 +1,32 @@
 name: root
 channels:
-  - defaults
-  - pyviz/label/dev
-  - bokeh/label/dev
-  - intake
   - conda-forge
 dependencies:
-  - jupyterhub==0.9.2
-  - bokeh=0.12.16
+  - jupyterhub
+  - bokeh
   - bqplot
+  - click
   - cython
   - cytoolz
-  - dask=0.18.2
+  - dask
   - dask-ml
+  - dask-kubernetes
   - datashader
-  - distributed=1.22.1
+  - distributed
   - esmpy
   - fastparquet
-  - gcsfs=0.1.2
+  - fusepy
+  - holoviews
+  - gcsfs
+  - geoviews
+  - hvplot
+  - intake::intake-xarray
   - ipywidgets
   - ipyleaflet
+  - jedi
   - jupyter
-  - jupyterlab=0.34.2
-  - jupyterlab_launcher=0.13.1
+  - jupyterlab=0.35
+  - jupyterlab_launcher
   - jupyter_client
   - holoviews
   - libsodium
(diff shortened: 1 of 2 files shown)


deployments/dev.pangeo.io/image/binder/environment.yml
```diff
@@ -1,62 +1,59 @@
...
+  - dask
   - dask-ml
+  - dask-kubernetes
   - datashader
-  - distributed=1.22.1
+  - distributed
   - esmpy
   - fastparquet
-  - gcsfs=0.1.2
+  - fusepy
```
mrocklin, line 18: do we still need fuse?

mrocklin (commented):
Looks fine to me.  @rsignell-usgs I also encouarge you to play things fast-and-loose here and feel comfortable self-merging things.

Generally I'm in favor of keeping explicit versions though.  Otherwise versions tend to shift about upstream, the cluster breaks, and folks don't know why.  This tends to increase maintenance burden on a few core folks.

We should definitely *update* some of these versions, but we should do so explicitly and it should probably be on the person doing the updating to verify that everything works nicely.

deployments/dev.pangeo.io/image/binder/environment.yml
rsignell-usgs: @mrocklin, I thought that by using only conda-forge channel, we wouldn't need to do all that pinning. 
(I have no idea whether we need fusepy.  I was just repeating what was there already.)

mrocklin: I suspect that the versions may still shift around in unpleasant ways.  Certainly having a consistent build chain helps to reduce conflicts due to different binary dependencies (like the GDAL pains of old) but versions of different libraries can still definitely shift away from each other.  Conda forge doesn't do any sort of ecosystem-wide testing.

rsignell-usgs: Okay, happy to "stand down" on this.  😸 
Would one strategy be to see if this works without pinning on "deploy", and if so, pin the versions on "staging"?

mrocklin: Yeah, I think it's entirely reasonable to unpin everything, deploy to a less-important deployment verify that things work and if so, then re-pin everything to those new versions.  Presuambly we would do this every few months.

rsignell-usgs: Great!  So okay to merge?

mrocklin: Assuming that you're ok passing through again and updating pinned versions then sure :)

rsignell-usgs: Yep!

PR #6, opened 2018-10-24
Retreat to pinned versions for dask

rsignell-usgs:
retreat to pinned versions for dask, distributed, dask-kubernetes and tornado


mrocklin:
I don't think that this was necessary.  Things worked fine for me on a
fresh install: https://gist.github.com/798bedf21e5b729561d77cec2d19144d

PR #13, opened 2018-10-25
replace any string with the correct image name

dsludwig:
@rsignell-usgs this should fix your issue with an out-of-date dask worker.

### DIFF
--- a/deployments/dev.pangeo.io/image/binder/start
+++ b/deployments/dev.pangeo.io/image/binder/start
@@ -8,7 +8,7 @@ export DASK_KUBERNETES__WORKER_NAME=dask-{JUPYTERHUB_USER}-{uuid}
 if [[ -z "${JUPYTER_IMAGE_SPEC}" ]]; then
     echo "JUPYTER_IMAGE_SPEC is not set"
 else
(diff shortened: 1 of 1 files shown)


jhamman:
Just to say, once https://github.com/dask/dask-kubernetes/pull/101 is merged, we wont need this `sed` business anymore.

mrocklin:
I'll take another look at that PR

rsignell-usgs:
I'm going to merge this, since @mrockin has encouraged me not to be shy on the develop branch!

rsignell-usgs:
@dsludwig I waited until the new container was built and stopped/started my server, but the container listed in `~/.dask/config.yaml` remains the same:   the original tag of `d03a8f7`.    I will manually change it to the latest tag `d2a0adb` from the list here: https://console.cloud.google.com/gcr/images/pangeo-181919/US/dev-pangeo-io-notebook?gcrImageListsize=50

dsludwig:
🤦‍♂️   this has the same issue as before - the `binder/start` file is not copied over, so it's still using the old file that doesn't have this change. I'll make an update to allow clobbering of home directory files.

PR #15, opened 2018-10-25
try using image env var

jhamman:
@rsignell-usgs and @dsludwig - I think we can try this.


rsignell-usgs:
Let's give it a try!

jhamman:
Seemed to work for me. @rsignell-usgs - how'd things work for you?

rsignell-usgs:
@jhamman, this worked for me also.

Issue #25, opened 2018-11-14
Use Google Cloud Filestore for shared storage


yuvipanda:
Google Cloud now has a managed NFS provider ([Filestore](https://cloud.google.com/filestore/)) that would be great for home directories. Currently, each user gets their own disk, which is expensive and rigid (you can't easily change sizes up or down after creation). It also makes some sharing scenarios harder.

Steps to use filestore:

1. Create a filestore 
2. Configure z2jh to use NFS as the backing store for home directories. We use one filestore for all users, and use `subPath` to give them rw access to each directory. This is used to scope users to directories, rather than traditional Unix user permissions
3. Use an initContainer to make sure the home directory created for the user has right permissions and ownership. See https://serverfault.com/questions/906083/how-to-mount-volume-with-specific-uid-in-kubernetes-pod for example.

https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/doc/source/amazon/efs_storage.rst is some info on doing something like this with EFS, which is used on AWS. Step (3) would be different. 

Once we have a good idea on how to set this up, this can be contributed back to the z2jh docs.

rabernat:
@raphaeldussin - we definitely want this for ocean.pangeo.io

yuvipanda:
Step 1: Convert default pool to use ubuntu for nodes, rather than default container optimized os. Ubuntu comes with ability to use NFS, this does not. I did this in the GUI in two steps:

1. Hit edit in the Kubernetes cluster page
2. Edit the nodepool
3. Turn 'node autorepair' to 'off'
4. Save

1. Hit edit in the Kubernetes cluster page
2. Edit the nodepool
3. Change 'node image' to ubuntu
4. Save

Progress is in the 'notifications' icon in top right. It takes a while and deletes / recreates all nodes. This disrupts all pods on the nodepool.

You can also specify this while creating the cluster.

yuvipanda:
Turns out this was a bad idea because of poddisruptionpolicy that is set by JupyterHub. I'm working around this manually just now, but in the future, the thing to do is to create a *new* nodepool with Ubuntu and kill the old nodepool after the new one is up.

yuvipanda:
This has been sorted out now. I'll continue tomorrow!

yuvipanda:
#28 is all the config required to make this happen. Woohoo!

I created the filestore instance with:

```
[...]

PR #26, opened 2018-11-15
Updating staging deployment

jhamman:
This clears the way for @yuvipanda to start work on #25.

@rsignell-usgs - Any objections here?


rsignell-usgs:
This is exciting!   No objections here!

PR #30, opened 2018-11-27
cartopy 0.17.0 => 0.16.0

rsignell-usgs:
cartopy 0.17.0 breaks https://github.com/jsignell/pydata_dc_2018/blob/master/04_panel.ipynb

### DIFF
--- a/deployments/dev.pangeo.io/image/binder/environment.yml
+++ b/deployments/dev.pangeo.io/image/binder/environment.yml
@@ -5,6 +5,7 @@ dependencies:
   - jupyterhub
   - bokeh
   - bqplot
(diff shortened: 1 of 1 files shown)


rabernat:
@rsignell-usgs -- I notice you making lots of updates to this dev cluster.

I am wondering if I can redirect some of your experience and effort to the nascent ocean.pangeo.io cluster being set up by @raphaeldussin. We are having quite a few challenges getting it working right:
https://github.com/raphaeldussin/example.pangeo.io-deploy/issues

We would be glad to incorporate your coastal ocean modeling work into this cluster, which we hope can serve the oceanographic community.

PR #45, opened 2019-02-07
bump repo2docker to 0.7.0 release and hubploy to yuvi-master

### LINKED ISSUE #40: Move to current master of hubploy
Currently, this is using hubploy from @dsludwig's fork. We've incorporated all the changes from the fork to hubploy master / repo2docker. We should try move this back to using hubploy master.

This should ideally happen at the same time as consolidating all hubs into one repo

jhamman:
closes #40


jhamman:
I'm going to try this and see how it works.

PR #49, opened 2019-02-07
hubploy master -- based on berkeley datahub setup

jhamman:


jhamman:
The PRs are building here now so that is good. This should be 99% of the way there for both the dev and ocean clusters.

PR #53, opened 2019-02-13
add seawater pkg

raphaeldussin:
@jhamman @rabernat  I added a package just to have something to commit.
Should we try to circle ci again?


rabernat:
Shouldn't we be using `gsw` and not `seawater`?

raphaeldussin:
I just needed something to push so we can try to re-run the ci with a PR with the right head ref.

pangeo-data/dev.pangeo.io-deploy#52

we can ditch seawater after. 

My circleci jobs failed on a permission denied. I guess it's good that one's PR do not change the
cluster state, how could we get some validation out of it ?

rabernat:
ok gotcha! merge away

PR #54, opened 2019-02-13
update version of repo2docker

jhamman:
@raphaeldussin - this should fix the problem we were seeing in repo2docker.


raphaeldussin:
@jhamman cool!

yuvipanda:
You can also do repo2docker master. A new version should hopefully be out later this week

PR #78, opened 2019-02-21
Turn HTTPS back on for ocean staging

yuvipanda:
DNS has been set.

### DIFF
--- a/deployments/ocean/config/staging.yaml
+++ b/deployments/ocean/config/staging.yaml
@@ -2,6 +2,6 @@ jupyterhub:
   proxy:
     https:
       hosts:
-      - ocean.pangeo.io
+      - staging.ocean.pangeo.io
       letsencrypt:
         contactEmail: raphael.dussin@gmail.com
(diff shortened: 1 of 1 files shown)


deployments/ocean/config/staging.yaml
```diff
@@ -1,7 +1,6 @@
 jupyterhub:
   proxy:
     https:
-      enabled: false
       hosts:
       - ocean.pangeo.io
```
rabernat: should this be `staging.ocean.pangeo.io`?

yuvipanda: @rabernat aha, yes it should be!

rabernat:
How do we ensure that the cluster stays at the same IP address? Does that need to go into the config?

yuvipanda:
@rabernat yes! You need to do two things:

1. Set `jupyterhub.proxy.service.loadBalancerIP` in staging to current IP
2. Go to https://console.cloud.google.com/networking/addresses/list?folder=&organizationId=&project=pangeo-181919 and mark that IP as 'static' - so we don't lose it without explicitly deleting it.

This would make sure that we have access to this IP forever, and lets us switch hubs between clusters if we need to.

PR #74, opened 2019-02-20
added nasa deployment folder

scottyhq:
this is for our first attempt at deploying pangeo jupyterhub via hubploy on AWS.

currently blocked by 
https://github.com/yuvipanda/hubploy/pull/14

PR #79, opened 2019-02-21
update readme

rabernat:
We need to update the readme to reflect the reality of how the clusters are built and deployed.

There is *one circle job* per branch. And we only have two branches: `staging`  and `prod`. These build and deploy all the clusters.

The clusters themselves are now listed in a separate table.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,17 +1,23 @@
-# About
-
-This repository contains the reproducible configuration for deploying a Pangeo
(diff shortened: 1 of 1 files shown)


rabernat:
Need to add the new AWS clusters as well.

jhamman:
We can add the AWS clusters in #74

PR #87, opened 2019-02-22
add NFS homeDirectories section to ocean

### LINKED ISSUE #84: NFS mounting issue
I am getting this in my event log as it tries to start my server

`2019-02-21 21:23:40+00:00 [Warning] MountVolume.SetUp failed for volume "ocean-staging-home-nfs" : mount failed: exit status 1 Mounting command: systemd-run Mounting arguments: --description=Kubernetes transient mount for /var/lib/kubelet/pods/a6af3b91-361e-11e9-94e8-42010a800086/volumes/kubernetes.io~nfs/ocean-staging-home-nfs --scope -- /home/kubernetes/containerized_mounter/mounter mount -t nfs <nil>:/ /var/lib/kubelet/pods/a6af3b91-361e-11e9-94e8-42010a800086/volumes/kubernetes.io~nfs/ocean-staging-home-nfs Output: Running scope as unit: run-r9a5ddc860e034ff9ac1e10447920c217.scope Mount failed: mount failed: exit status 32 Mounting command: chroot Mounting arguments: [/home/kubernetes/containerized_mounter/rootfs mount -t nfs <nil>:/ /var/lib/kubelet/pods/a6af3b91-361e-11e9-94e8-42010a800086/volumes/kubernetes.io~nfs/ocean-staging-home-nfs] Output: mount.nfs: Failed to resolve server <nil>: Name or service not known`

rabernat:
This fixes #84.

In this PR, I have chosen the path on the nfs server called `home/pangeo-users/{username}` in which to place the user home directories. The idea is that **the same directory will be used on all clusters from the same region**. I chose this because, I personally plan to log in all these clusters, and I would like to see the same home directory.

I would like some feedback on whether this is the right choice.

### DIFF
--- a/deployments/dev/config/common.yaml
+++ b/deployments/dev/config/common.yaml
@@ -9,12 +9,12 @@ jupyterhub:
         volumeMounts:
         - name: home
           mountPath: /home/jovyan
-          subPath: "home/dev.pangeo.io/{username}"
+          subPath: "home/pangeo-users/{username}"
     storage:
       type: static
       static:
         pvcName: home-nfs
-        subPath: "home/dev.pangeo.io/{username}"
+        subPath: "home/pangeo-users/{username}"
     cloudMetadata:
       enabled: true
     cpu:
--- a/deployments/ocean/config/common.yaml
+++ b/deployments/ocean/config/common.yaml
@@ -9,12 +9,12 @@ jupyterhub:
         volumeMounts:
         - name: home
           mountPath: /home/jovyan
-          subPath: "home/ocean.pangeo.io/{username}"
+          subPath: "home/pangeo-users/{username}"
     storage:
       type: static
       static:
         pvcName: home-nfs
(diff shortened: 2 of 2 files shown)


rabernat:
@jhamman in response to your q in #84, I updated this PR to use the same NFS mount point on the dev cluster. That will let us test how it works.

jhamman:
Okay, sounds good. I'm just curious what you think the expected behavior will be when there are conflicts in the home space. Feel free to merge this so we can test in anger.

rabernat:
> I'm just curious what you think the expected behavior will be when there are conflicts in the home space.

I honestly don't understand how this all works enough in order to make a prediction. Could you explain what you mean, since you obviously have some thoughts here?

How is the home space populated?

jhamman:
I'm expecting that there will be conflicts when new images mount the same home dir space. Let's find out...

PR #90, opened 2019-02-27
Depend on pangeo chart

jhamman:


jhamman:
Made possible by https://github.com/yuvipanda/hubploy/commit/7dab0db22f7c79aef01c4c73e798764318817a70

pangeo-deploy/values.yaml
```diff
@@ -7,34 +7,34 @@
...
+        tag: 2019.02.01
+      cpu:
+        limit: 2
+        guarantee: 1
+      memory:
+        limit: 4G
+        guarantee: 2G
+      defaultUrl: "/lab"
+      cmd: jupyter-labhub
+      serviceAccountName: daskkubernetes
```
jhamman, line 33: I think we can remove basically all of this.

yuvipanda: Because it's present in the pangeo chart?

yuvipanda:
You need to set images.image_config_path in hubploy.yaml as well. LMK how this goes!

rabernat:
Awesome Joe! Thanks for this!

PR #95, opened 2019-03-02
make builds always run

rabernat:
This should make it so all clusters always try to build / deploy regardless of the exit status of the others.


rabernat:
worked

PR #129, opened 2019-03-05
add dask config

jhamman:


jhamman:
the build passed but the config file copy failed. AKA don't merge yet.

PR #143, opened 2019-03-07
add admin users

rabernat:
adds this block under globus section
```
        admin:
          access: true
          users:
            - 0000-0001-7479-8439 # Joe Hamman
            - 0000-0001-5999-4917 # Ryan Abernathey
            - 0000-0003-4004-4553 # Raphael Dussin
```


rabernat:
For some reason, this broke authentication.

PR #144, opened 2019-03-07
Revert "add admin users"

rabernat:
Reverts pangeo-data/pangeo-cloud-federation#143

Still need to figure out how to add admin users with globus.


rabernat:
false alarm

PR #150, opened 2019-03-09
Clean up ocean image

rabernat:
I am trying something weird here which is replacing the ocean image with a submodule that points to the pangeo_ocean_examples binder.


rabernat:
This didn't work because the submodule at `deployments/ocean/image` did not get cloned. Need to update circleci script.

PR #186, opened 2019-04-01
tried using emptydir instead of gitrepo for templates

scottyhq:


jhamman:
@scottyhq - should we merge this and give it a try?

PR #199, opened 2019-04-04
production auth for icesat2

scottyhq:


deployments/icesat2/config/staging.yaml
```diff
@@ -4,7 +4,7 @@ pangeo:
     proxy:
       https:
         hosts:
-        - staging.icesat2.pangeo.io
+        - icesat2.pangeo.io
```
jhamman: this change should be in .../config/prod.yaml.

jhamman:
I think we still need to `deployments/icesat2/secrets/prod.yaml` file. There will be some auth secrets in there.

deployments/icesat2/config/staging.yaml
scottyhq: i think we're good to go now

PR #223, opened 2019-04-19
update ocean image

rabernat:
Is this still the right way to point to the docker image? Or can our new "base image" magic help us with this somehow?

### DIFF
--- a/deployments/ocean/image/binder/Dockerfile
+++ b/deployments/ocean/image/binder/Dockerfile
@@ -1,2 +1,2 @@
 # Note that there must be a tag
-FROM pangeo/pangeo-notebook:2019.04.01
+FROM pangeo/pangeo-notebook:2019.04.19


jhamman:
@rabernat - this is fine. You can also you the `onbuild` version of the image to start from a previous image and still install new packages on top. All the normal repo2docker rules would apply. In other words, you could just add a environment.yml or postBuild file, in addition to the docker file and new packages would be added on top of the existing image.

yuvipanda:
Yep, this would work. 

Would love to use the onbuild variants for adding ocean specific packages though. It can be used to experiment here and then be made its own image in Pangea stacks

rabernat:
Ironically we recently just removed the pangeo-ocean image from pangeo stack because we deemed it was unnecessary fragmentation: https://github.com/pangeo-data/pangeo-stacks/pull/24

IMO, what we want is a pretty comprehensive image like we have now in pangeo-notebook, plus the ability for individual hub users to add packages they need without assistance from the hub admins. We create toil by requiring hub admins to respond to requests for new package installs.

rabernat:
I will try the `onbuild` suggestion in the future. For now I have science to do and I need tensorflow.

PR #252, opened 2019-04-29
first pass at getting dask workers to share user home directory

scottyhq:
@yuvipanda @jhamman - i'm getting fairly confused with PVs, PVCs, volumes, NFS, EFS. But this seems like it should accomplish dask workers having access to a users home directory. 

Reminder - this needed to share custom conda environments and (maybe?) a users credentials in ~/.aws


yuvipanda:
Yeah, something along these lines (assuming these are the same volume and volumeMount bits that the user pod has) should work.

Issue #148, opened 2019-03-07
dask labextension not working in ocean


rabernat:
I am seeing this in my notebook pod logs:

```
404 GET /user/0000-0001-5999-4917/dask/clusters?1551920957641 (0000-0001-5999-4917@10.128.0.18) 2.41ms
```

after I click on the button to create a new cluster.

ian-r-rose:
@rabernat Is it possible that the server extension is not installed or enabled? The newer version of the extension requires it. We also haven't hammered out all the kinks in usage with `dask-kubernetes` yet, so I'm quite curious to see how this works.

In a terminal:
```bash
jupyter serverextension list
```

jhamman:
Just ran that on the ocean deployment:

```
$ jupyter serverextension list
config dir: /srv/conda/etc/jupyter
    jupyterlab  enabled
    - Validating...
      jupyterlab 0.35.4 OK
    nteract_on_jupyter  enabled
    - Validating...
Error loading server extension nteract_on_jupyter
      X is nteract_on_jupyter importable?
    nbserverproxy  enabled
    - Validating...
      nbserverproxy  OK
```

ian-r-rose:
Is the nasa deployment working? I see that it includes the python package, while the ocean deployment does not: https://github.com/pangeo-data/pangeo-cloud-federation/blob/staging/deployments/nasa/image/binder/environment.yml#L39

rabernat:
So in the latest configuration, the dask labextension works, but it launches LocalCluster, rather than kubeCluster, by default.

I think all that is needed is an extra piece of dask config. So the questions are:
- What is this config?
- Where is the best way to inject it into the environment. (The notebook image is just a straight copy of `pangeo-notebook:2019-04-19` from pangeo-stacks.)

jhamman:
@rabernat - you may want to look at the NASA deployment. 

- we have the dask lab extension set as: https://github.com/pangeo-data/pangeo-cloud-federation/blob/1bbab64ec0c4dd14eb18c74fd7610e0f6f7ecd68/deployments/nasa/image/binder/dask_config.yaml#L39-L44
- we overwrite the default config here: https://github.com/pangeo-data/pangeo-cloud-federation/blob/1bbab64ec0c4dd14eb18c74fd7610e0f6f7ecd68/deployments/nasa/image/binder/postBuild#L14

ian-r-rose:
[...]

PR #258, opened 2019-05-03
try copying dask config

rabernat:
Yet another possible fix for #148.


yuvipanda:
This is the right thing to do!

PR #267, opened 2019-05-07
265, 266 to prod

jhamman:
This removes node selectors from GKE clusters


jhamman:
failed build was just a timeout on circleci. merging...

PR #271, opened 2019-05-07
#268, #269, #270 to prod

jhamman:


jhamman:
The failing test is a timeout in the nasa deploying and is not related to these changes. Merging.

PR #279, opened 2019-05-14
Per namespace shared nfs for nasa hubs

scottyhq:
this adds /home/jovyan/shared to both staging and prod hubs. It is just the config files though. requires running manually, not currently incorporated into hubploy as with user home directories on nfs


jhamman:
@scottyhq - is this ready to go?

Issue #276, opened 2019-05-12
dev.pangeo.io - Only 51% of worker nodes utilized


consideRatio:
I found in the cloud console that we have a lot of nodes that are underutilized, because they have pods scheduled on them that request 50% of the nodes memory. I bet the person scheduling these pods aim to get two pods per node but is actually ending up with only one pod per node by mistake.

![image](https://user-images.githubusercontent.com/3837114/57578693-80b5f880-7491-11e9-9272-e32e60854f6e.png)

Inspecting the node further, one can see that there are some system pods scheduled on each node related to logging and such, these will always eat up some resources.

![image](https://user-images.githubusercontent.com/3837114/57578716-edc98e00-7491-11e9-9c76-c7b1fdf43f0b.png)

Reducing the request to 11.5 would solve this. I'm currently too fresh on this code base to figure out what to do with this insight so I figured I'd write it here.

jhamman:
Thanks for bringing this up @consideRatio. I think we have a few things we can do here. 

1. Finish up the dask-kubernetes taints/affinity PRs (https://github.com/dask/dask-kubernetes/pull/147)
2. Come up with a way to document/publish k8s cluster configurations to help users better determine appropriate pod specs. 
3. Work on getting grafana dashboards setup for these clusters so people can see the cluster usage without being an admin on the GCP account (see https://github.com/pangeo-data/helm-chart/pull/88)
4. Experiment with GKE's [node auto-provisioning](https://cloud.google.com/kubernetes-engine/docs/how-to/node-auto-provisioning) for dask-workers.

I think, by chance, I've ordered these appropriately, in terms of both ease and relevance. What do you think?

consideRatio:
@jhamman that sounds very practical and suitable!

jhamman:
@rabernat - can we get you (or @raphaeldussin) to reduce the memory requirement in the dask-worker config for ocean.pangeo.io: https://github.com/pangeo-data/pangeo-cloud-federation/blob/8f7f4bf9963ef1ed180dd20c952ff1aa8df54ca2/deployments/ocean/image/binder/dask_config.yaml#L32

11.5G would be fine. You might as well also ask for more cpu (1.75?) since there is more to be had on these nodes.

consideRatio:
@jhamman I think it is often best to let the request be quite low and the limit quite high for CPU, while you request/limit similarly for memory.

It makes it easier to accomplish max CPU utilization.

[...]

PR #280, opened 2019-05-14
reduce dask worker memory request on ocean

rabernat:
Possible fix for #276.

cc @consideRatio

### DIFF
--- a/deployments/ocean/image/binder/dask_config.yaml
+++ b/deployments/ocean/image/binder/dask_config.yaml
@@ -28,10 +28,10 @@ kubernetes:
         name: dask-${JUPYTERHUB_USER}
         resources:
           limits:
-            cpu: "1.75"
+            cpu: 16
             memory: "11.5G"
           requests:
-            cpu: "1.75"
+            cpu: 1
             memory: "11.5G"
 
 labextension:
--- a/deployments/ocean/image/binder/dask_config.yaml
+++ b/deployments/ocean/image/binder/dask_config.yaml
@@ -21,7 +21,7 @@ kubernetes:
           - '2'
           - --no-bokeh
           - --memory-limit
-          - 12GB
+          - "11.5GB"
           - --death-timeout
           - '60'
         image: ${JUPYTER_IMAGE_SPEC}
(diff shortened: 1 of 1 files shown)


deployments/ocean/image/binder/dask_config.yaml
```diff
@@ -21,18 +21,18 @@ kubernetes:
...
           - --memory-limit
-          - 12GB
+          - "11.5GB"
           - --death-timeout
           - '60'
         image: ${JUPYTER_IMAGE_SPEC}
         name: dask-${JUPYTERHUB_USER}
         resources:
           limits:
             cpu: "1.75"
```
consideRatio: I think this could be allowed to be be as high as the node you schedule on can use. It is an optimization without drawbacks as I understand it.

deployments/ocean/image/binder/dask_config.yaml
```diff
@@ -21,18 +21,18 @@ kubernetes:
...
-            memory: 12G
+            memory: "11.5G"
           requests:
-            cpu: 1
-            memory: 12G
+            cpu: "1.75"
```
consideRatio: I think it would make sense to lower this, as if you have two pods on a node and one is inactive currently, the other could if needed get more power by not making the pod hog a fixed amount that they wont share.

I also think this will be an optimization without drawbacks, and perhaps no real benefit either depending on the kind of work that is done by the worker.

deployments/ocean/image/binder/dask_config.yaml
rabernat: I'm a bit confused about the interaction between dask and kubernetes here. Each has its own independent mechanisms to manage CPU and memory usage. The dask worker process here is limited to 2 threads. It has no notion of "requests" vs "limits". 

In any case, i will try following your advice.

consideRatio: Ah I've been very uncertain about how a dask-worker consume resources. Is it correct that since since dask the dask workers are limited to 2 threads, they will never use more than 2 CPU?

Even though they are managed by linux cgroups etc, I think of these Kubernetes settings as restricting the pods resources and that they can be seen as a description of the dask-workers CPU/Memory usage restrictions, but they will also be used to influence how many dask workers can be fit onto a node and how much of the nodes total CPU/Memory will end up being in active use.

PR #282, opened 2019-05-15
fix conda config for ocean

### LINKED ISSUE #272: pip installs are not persistent on ocean
If I pip install a package on the ocean cluster, then shut down my notebook pod, it's not there when I start it back up.

What config needs to be changed to overcome this?

rabernat:
Fixes #272 by using custom conda config file.


scottyhq:
@rabernat you'll have to copy .condarc to /home/jovyan in common.yaml as discussed in this issue https://github.com/pangeo-data/pangeo-cloud-federation/issues/272. But i'd say leave this pull request open for a little bit until we hear about alternative approaches.... 

As a work around in the meantime existing users can manually put that .condarc in their home folder.

Issue #273, opened 2019-05-09
Scheduling dask worker pods


consideRatio:
I'm quite new in the domain of pangeo and Dask, but my understanding is that the users of a pangeo jupyter pod has the rights to use dask-kubernetes to spawn dask-worker pods. I created this issue to discuss the scheduling of these pods in order to downscale properly.

It is my guess, that when these schedule, they will use the default scheduler. The default scheduler will spread the pods across the available nodes. But, you probably want to pack them tight as that allows nodes to free up and scale down.

So, I'm proposing to consider using the Z2JH helm chart's provided scheduler instead.

### Example when packing pods would be important

Say that you have 100 users and they all schedule 10 pods each, and this leads to 20 powerful nodes (dedicated for worker pods) are created for example. These 100 users now all leave at the same time and suddenly there are no worker pods scheduled on these 20 powerful nodes. But then two users drop in, they both schedule 10 worker pods that are scheduled by the default scheduler which spreads them apart on the 20 nodes. Suddenly 20 nodes are utilized and may therefore not scale down...

If the z2jh user pod scheduler would been used instead, it would have packed the pods tight on one node instead, making 19 other nodes scale down keeping one running only for the two users.

### Concerns of this idea
It may require the z2jh provided scheduler to schedule pods in different namespaces, is it configured for this? I don't remember. But we could probably make this configurable in the z2jh helm chart if not.

I tried evaluating this right away, but I'm still not confident about the situation. But now I know that the z2jh user pod scheduler has rights to list and schedule pods in all namespaces, but is its configuration making it available to do so? I think so but I'm not sure.

#### Related to the namespace concern
- Anything in the the [z2jh user-scheduler folder](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/jupyterhub/templates/scheduling/user-scheduler)
- The z2jh user pod scheduler's [container initialization](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/eaf87a217fca1834e299a0567a1ef87d813369b7/jupyterhub/templates/scheduling/user-scheduler/deployment.yaml#L30-L41)
- The [RBAC permissions](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/eaf87a217fca1834e299a0567a1ef87d813369b7/jupyterhub/templates/scheduling/user-scheduler/rbac.yaml#L10-L24) of the z2jh user pod scheduler

jhamman:
[...]

PR #283, opened 2019-05-16
turn on userScheduler again

jhamman:
xref: #273

@consideRatio - do you think we need the userPlaceholder feature at this point?

### DIFF
--- a/pangeo-deploy/values.yaml
+++ b/pangeo-deploy/values.yaml
@@ -42,8 +42,7 @@ pangeo:
       podPriority:
         enabled: true
       userPlaceholder:
-        enabled: true
-        replicas: 2
+        enabled: false
 
 homeDirectories:
   nfs:
--- a/pangeo-deploy/values.yaml
+++ b/pangeo-deploy/values.yaml
@@ -37,8 +37,13 @@ pangeo:
         enabled: false
 
     scheduling:
-        userPlaceholder:
-          enabled: false
+      userScheduler:
+        enabled: true
+      podPriority:
+        enabled: true
+      userPlaceholder:
+        enabled: true
+        replicas: 2
 
 homeDirectories:
   nfs:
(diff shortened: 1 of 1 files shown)


consideRatio:
Hmmm in short, for both a good up (user placeholders) and down (user scheduler) scaling experience, enable at least one user placeholder and pod priority as well. For more details see the [optimization section](http://z2jh.jupyter.org/en/latest/optimization.html) of the z2jh guide for details.

- If you don't have the user scheduler enabled, you won't get nodes packed tight, which really matters in order to help with downscaling.
- If you don't have user placeholders enabled and >0 replicas of them, you won't scale up the jupyter user nodes ahead of time, that will user need to wait for example 4-8 minutes when they scale up.
- NOTE: enabling user-placeholder pods requires you to enable podpriority as well, and that should only be done on k8s 1.11+. Furthermore, some technical details may differ in between cloud provider at least before k8s 1.13.

consideRatio:
LGTM @jhamman !

If you have non GKE clusters, I'd verify the following regarding the podPriority setting, that is not a given to work on all different K8s clusters of versions that isn't super fresh, but will work great on GKE at least. I know little about AWS / Azure etc. though.

1. What is the priority given pods in general?
2. What is the priority given to of user-placeholder pods?
3. What is the priority cutoff of the cluster? This will depend and be hard to inspect, so I'd simply try verifying that it is possible to evict user-placeholder pods with normal users. One way to see if this happens is to see that they first manage to schedule somewhere, and then becomes pending, and then schedules on another node and get their running time reset.

consideRatio:
Disclaimer and a rule of thumb I conjure out of the air: having user placeholders is probably a bad idea if you want to scale down to 0 nodes, but a good idea if you don't.

jhamman:
> Disclaimer and a rule of thumb I conjure out of the air: having user placeholders is probably a bad idea if you want to scale down to 0 nodes, but a good idea if you don't.

Yes, we do want to scale down to zero nodes (for now). That's why I was previously disabling this feature.

jhamman:
@scottyhq - can you remind me what version of k8s the two aws clusters are running?

PR #303, opened 2019-06-18
adding h5pyd current github master installed via pip

rsignell-usgs:
I'd like to add h5pyd master from github so I can test the HSDS service vs zarr with kubernetes

### DIFF
--- a/deployments/icesat2/image/binder/environment.yml
+++ b/deployments/icesat2/image/binder/environment.yml
@@ -46,11 +46,13 @@ dependencies:
   - pandas=0.24
   - geopandas=0.5
   - rasterio=1.0
+  - pip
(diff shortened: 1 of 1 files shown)


rsignell-usgs:
@scottyhq I tried stopping and restarting my service on ice2, and the `h5pyd` we tried to add is not there.  I see that circleci passed, but the comment says:
```
Image 783380859522.dkr.ecr.us-west-2.amazonaws.com/pangeo:71136292559c: already up to date.
```

Does that mean the new docker container did not get built for some reason?

rsignell-usgs:
@scottyhq nevermind.  I was not on `staging.ice2`.    `h5pyd` is there on staging and everything is fine.

PR #305, opened 2019-06-21
avoid pip

ocefpaf:
pip installing in a conda env may clobber some of the dependencies installed with conda and cause problems. (Sometimes things just works but it is 50-50 change.)

PS: In a future PR I'm planning on adding the `defaults` channel below conda-forge here. In the past we did recommend removing `defaults` but since conda >=4.6 we have a better strategy to handle binary incompatibility issues with the `strict` channel option. (Environments are more reliable and stable that way.)

### DIFF
--- a/deployments/ocean/image/binder/environment.yml
+++ b/deployments/ocean/image/binder/environment.yml
@@ -2,6 +2,5 @@ name: pangeo
 channels:
   - conda-forge
 dependencies:
-  - pip:
-    - dask-labextension==0.3.1
-    - nbgitpuller
+  - dask-labextension==0.3.1
+  - nbgitpuller


rsignell-usgs:
@scottyhq, no issues on merging this I hope!

ocefpaf:
BTW we (conda-forge) are working on making the packages smaller to get lighter docker images. See https://jcrist.github.io/conda-docker-tips.html for some of the results.

rsignell-usgs:
@ocefpaf, whoa, that's a massive reduction in container size!  I read it first as a 15% reduction, but it's 15% of the original size!   So a more than a 5-fold reduction in container size!

scottyhq:
this is great, thanks @ocefpaf ! I think we should implement this in pangeo-stacks
(https://github.com/pangeo-data/pangeo-stacks/pull/53)

and then inherit the changes in each deployment, as we do here:
https://github.com/pangeo-data/pangeo-cloud-federation/blob/fc5ace29eb6fc66c842bdc05035b0e4d9aac48b7/deployments/nasa/image/binder/Dockerfile

ocefpaf:
> this is great, thanks @ocefpaf ! I think we should implement this in pangeo-stacks (pangeo-data/pangeo-stacks#53)

By this you mean the pip part or the other suggestions from that post?

> and then inherit the changes in each deployment, as we do here:
/deployments/nasa/image/binder/Dockerfile@fc5ace2

I'm new here so I'm lost with all the repos and folders. Can you point me to the canonical/first Dockerimage? I'm happy to port the post suggestions and some other conda improvements there.

jhamman:
@ocefpaf - the base docker file used here (and in pangeo-stacks) comes from repo2docker. How do you turn on the `strict` option? Repo2docker uses a shell command to setup conda (see here: https://github.com/jupyter/repo2docker/blob/cd809645a9e00cf14abd110d8fb8098193379a7c/repo2docker/buildpacks/conda/install-miniconda.bash#L31-L36). Unfortunately, its not easy to modify the behavior of conda in this framework but with a bit more information from you on the desired configuration path, I think we can come up with something.

scottyhq:
The pip part for starters we can take care of.

As for the Dockerfile suggestions (https://jcrist.github.io/conda-docker-tips.html), in addition to what @jhamman said I'll point out that the base image is Ubuntu
https://github.com/jupyter/repo2docker/blob/5e67e3c74362bfa9a8e7f457f406ab4af64e3697/docs/source/design.md#reproducibility-and-version-stability 

Pinging @yuvipanda on this and adding a link to a relevant issue about different base images with repo2docker https://github.com/jupyter/repo2docker/issues/487

ocefpaf:
> @ocefpaf - the base docker file used here (and in pangeo-stacks) comes from repo2docker. How do you turn on the `strict` option? Repo2docker uses a shell command to setup conda (see here: [jupyter/repo2docker:repo2docker/buildpacks/conda/install-miniconda.bash@`cd80964`#L31-L36](https://github.com/jupyter/repo2docker/blob/cd809645a9e00cf14abd110d8fb8098193379a7c/repo2docker/buildpacks/conda/install-miniconda.bash#L31-L36)). Unfortunately, its not easy to modify the behavior of conda in this framework but with a bit more information from you on the desired configuration path, I think we can come up with something.

See https://github.com/jupyter/repo2docker/pull/714, but we are planning on pushing for a feature request where we can enable `strict` in the `environment.yaml` file, which will make this change explicitly and visible where it matters the most.

jhamman:
@ocefpaf - thanks for opening this. We've made progress upstream and now are only using pip to install dev versions of a few packages. I'm going to close this since your changes are now pretty far out of date. 🍻

PR #313, opened 2019-06-26
Add OOI deployment (using staging branch)

tjcrone:
It seems that if I make a pull request from another branch into staging, the entire CircleCI workflow does not run. Not sure if I am doing something wrong.


tjcrone:
Actually I think the only way that the deploy job runs is if I merge this. That makes sense.

PR #319, opened 2019-07-03
bump helm chart version

jhamman:


TomAugspurger:
Thanks. Just to make sure, this isn't likely to solve https://github.com/pangeo-data/pangeo-binder/issues/59 right? This is just your xarray bump? I'm looking into that in a few minutes.

And are these changes continuously deployed, or is there a manual deploy step?

jhamman:
@TomAugspurger - 

#318 was the xarray version bump. That is live on the staging namespaces now. This PR bumps the version of z2jh. I don't expect this to change anything but we were 3 months stale so I'm just keeping things fresh.

I bet pangeo-data/pangeo-binder#59 is related to jupyterlab 1.0. Just a guess though.

These changes will be continuously deployed but not to binder. That happens in the pangeo-binder repo.

PR #324, opened 2019-07-14
initial esip config

scottyhq:
@rsignell-usgs and @jhamman


scottyhq:
going to take the liberty of merging this to test the deployment out. maybe in the future we should set up some sort of required review...

PR #330, opened 2019-07-17
Updating ocean environment.yml file to include xmitgcm

shanicetbailey:
Trying to include xmitgcm package so that we can use llcreader module for ECCO cloud datasets.


rabernat:
Thanks Shanice!

As long as we are updating this, are there any other new packages we want to include in this PR? Xhistogram? 

cc @cspencerjones, @roxyboy, @dhruvbalwada

PR #338, opened 2019-07-26
Update requirements.txt and other tools used in circleci workflow

scottyhq:
most important change here is moving latest release of repo2docker into requirements.txt instead of .circleci/config.yml. Some AWS version bumps and a helm version bump (which might require updating tiller version on clusters). Thoughts @jhamman and @tjcrone ?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,9 +29,6 @@ jobs:
             python3 -m venv venv
             source venv/bin/activate
             pip install --upgrade -r requirements.txt
-            # See https://github.com/pangeo-data/pangeo-cloud-federation/issues/120
-            # Until https://github.com/jupyter/repo2docker/pull/605
-            pip install --upgrade git+https://github.com/yuvipanda/repo2docker@d040bfa57addd7252afbfe0a4863f119bca9ab8a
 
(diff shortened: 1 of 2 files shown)


jhamman:
This looks fine to me. I guess the only way to test this out is to deploy to staging.

tjcrone:
This looks great. I am going to merge because I would like to try these changes during my testing today. Thanks Scott!

tjcrone:
Actually just noticed the [WIP] tag. So I will not merge without more input from Scott. But looks good to me!

scottyhq:
yeah, i'm in favor of deploying to staging. we recently ran into mismatched version errors for helm with the esip deployment (see https://stackoverflow.com/questions/50701224/helm-incompatible-versions-between-client-and-server). 

So https://github.com/pangeo-data/pangeo-cloud-federation/blob/6b54c8b854c4987a5a6ac05cca6afbe0fac759d5/.circleci/config.yml#L237 could be changed to `helm init --upgrade` to ensure the same versions. Or we leave it up to each cluster deployer to upgrade tiller if need be?

tjcrone:
Yeah I still think the best solution to this problem might be to go with separate branches for each deployment so they can be changed and redeployed independently. Or maybe there is some way that we can coax CircleCI into knowing which deployments changed and thus know which deployments to upgrade. This sounds doable but I haven't thought deeply about it. We can't be the first group to have run into this particular issue. Has anyone heard of or thought of other potential solutions?

PR #369, opened 2019-08-15
Merge esip staging into prog

rsignell-usgs:


scottyhq:
minor change to nasa prod hub, so fine by me @rsignell-usgs !

PR #372, opened 2019-08-22
Add erddapy to ocean image

tjcrone:
This pull request fixes a minor typo and also adds [erddapy](https://ioos.github.io/erddapy) to the ocean.pangeo.io image.

cc @ocefpaf, @emiliom

PR #373, opened 2019-08-22
Packages that will be used in the OHW19

ocefpaf:
Follow up from #372. This PR adds the extra packages that we will be using in the OHW19. Please let me know if this is the right file we should be modifying.

### DIFF
--- a/deployments/ocean/image/binder/environment.yml
+++ b/deployments/ocean/image/binder/environment.yml
@@ -18,6 +18,7 @@ dependencies:
   - owslib
   - palettable
   - pocean-core
+  - podaacpy
   - retrying
   - seaborn
   - sympy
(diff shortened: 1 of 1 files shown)


ocefpaf:
Ping @rabernat and @tjcrone.

deployments/ocean/image/binder/environment.yml
```diff
@@ -18,6 +18,7 @@ dependencies:
   - owslib
   - palettable
   - pocean-core
+  - podaacpy
```
ocefpaf, line 21: Ping @emiliom.

tjcrone:
@ocefpaf, this looks great. I talked to Joe and he said we are fine to merge to staging during the Pangeo conference. One question, did you confirm that all of these packages are not already installed?

tjcrone:
A quick check seemed to indicate that owslib might already be there? Might be best not to install again?

ocefpaf:
> One question, did you confirm that all of these packages are not already installed?

I did but as you noticed I overlooked `owslib`. Just rebased against the latest commit on `staging` and removed `owslib`.

PR #375, opened 2019-08-22
Update ooi whitelist

lsetiawan:
Adding myself to OOI whitelist since org whitelist doesn't work.

### DIFF
--- a/deployments/ooi/config/common.yaml
+++ b/deployments/ooi/config/common.yaml
@@ -119,6 +119,7 @@ pangeo:
           - rabernat
           - scottyhq
           - yuvipanda
+          - lsetiawan
       admin:
         access: true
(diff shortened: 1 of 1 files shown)


lsetiawan:
ping @tjcrone This is needed because you're using people whitelist, not just organization. I think to change this behaviour this config needs to be updated so that you don't use user whitelist. Thanks!

tjcrone:
@lsetiawan, are you saying that it is not possible to mix a user whitelist with the organization whitelist?

tjcrone:
If this is the case, can you please instead update this PR so that you remove the user whitelist section entirely?

lsetiawan:
@tjcrone I have updated with the config suggested.

PR #394, opened 2019-09-03
bump some icesat2 env versions

jhamman:
cc @scottyhq

### DIFF
--- a/deployments/icesat2/image/binder/environment.yml
+++ b/deployments/icesat2/image/binder/environment.yml
@@ -16,18 +16,19 @@ dependencies:
   - nomkl
   - s3fs=0.2
   - gcsfs=0.2
-  - intake=0.4
+  - intake=0.5
   - intake-xarray=0.3
   - msgpack-python=0.6
   - curl=7.64
   - python=3.6
   - xarray=0.12
-  - zarr=2.2
+  - nc-time-axis=1.2
+  - zarr=2.3.2
   - h5py=2.9
   - h5netcdf=0.7
-  - dask=1.1
-  - distributed=1.25
-  - dask-kubernetes=0.8
+  - dask=2.3
+  - distributed=2.3
+  - dask-kubernetes=0.9.2
   - matplotlib=3.1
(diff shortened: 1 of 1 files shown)


deployments/icesat2/image/binder/environment.yml
```diff
@@ -16,18 +16,19 @@ dependencies:
...
   - h5py=2.9
   - h5netcdf=0.7
-  - dask=1.1
-  - distributed=1.25
-  - dask-kubernetes=0.8
+  - dask=2.3
+  - distributed=2.3
+  - dask-kubernetes=0.9.2
   - matplotlib=3.1
   - nbserverproxy=0.8
```
andersy005, line 33: `nbserverproxy` got rename to `jupyter-server-proxy`

deployments/icesat2/image/binder/environment.yml
```diff
@@ -16,18 +16,19 @@ dependencies:
   - nomkl
   - s3fs=0.2
```
andersy005, line 17: There's a more recent release of s3fs too

deployments/icesat2/image/binder/environment.yml
```diff
@@ -16,18 +16,19 @@ dependencies:
...
   - gcsfs=0.2
-  - intake=0.4
+  - intake=0.5
```
andersy005, line 19: Do we need to specify `intake` as a dependency? I think that having `intake-xarray` takes care of this

scottyhq:
fine by me, but before merging this let's add a profile list so that hackweek attendees can checkout the image with an exact copy of packages used during the hackweek. See (https://github.com/pangeo-data/pangeo-cloud-federation/blob/70580aa714b4a4b9d5375467f98f5c7e92ebe5ca/deployments/nasa/config/common.yaml#L61). We can point to the image that I copied over to DockerHub - `uwhackweeks/icesat2:2019`.

jhamman:
@scottyhq - I have to run. Can you both make those changes in separate PRs?

Issue #398, opened 2019-09-04
icesat2 hub not scaling up


jhamman:
Following some changes to the icesat2 profile list yesterday, we're having some problems getting user pods to schedule:

```
Event log
Server requested
2019-09-04 16:51:28+00:00 [Warning] 0/3 nodes are available: 3 node(s) didn't match node selector.
2019-09-04 16:51:38+00:00 [Normal] pod didn't trigger scale-up (it wouldn't fit if a new node is added)
Spawn failed: Timeout
```

@scottyhq - can you take a look at the logs to see if anything pops out at you? Are the pods in the profile list too big?

scottyhq:
sure enough @jhamman, the icesat2 cluster needs some maintenance. the nodes have been up for 100+ days and the original configuration is for 20Gb disks. As far as I know Docker images remain if a node isn't shut down and our current configuration is messy, with 1 user-notebook node running all the time:

```
CLUSTER	NODEGROUP	CREATED			MIN SIZE	MAX SIZE	DESIRED CAPACITY	INSTANCE TYPE	IMAGE ID
pangeo	dask-worker	2019-04-09T16:30:58Z	0		20		0					
pangeo	ng-680864af	2019-03-28T23:07:24Z	3		3		3					
pangeo	test		2019-04-05T21:36:33Z	0		100		1					
pangeo	user-notebook	2019-04-09T16:38:17Z	0		20		0
```

autoscaler log:
```
Pod can't be scheduled on eksctl-pangeo-nodegroup-user-notebook-NodeGroup-XXFTQEPOCHSU, predicate failed: CheckNodeDiskPressure predicate mismatch, cannot put icesat2-staging/jupyter-scottyhq on template-node-for-eksctl-pangeo-nodegroup-user-notebook-NodeGroup-XXFTQEPOCHSU-2143810545988417440, reason: node(s) had disk pressure
```

I suspect the autoscaler sees there is already a node up and instead of launching a new one prefers to schedule there, but then runs into the disk pressure issue.

We should update the nodegroups to match the NASA deployment (I should be able to get to it this week)

scottyhq:
Fixed by #401 and #402. We now use 100Gb EBS volumes on nodes and the icesat2 and nasa cluster configurations are the same.

Issue #356, opened 2019-07-30
aws-node NetworkPlugin cni failures on AWS Hubs


scottyhq:
For some time we've been noticing network related errors on AWS hubs when scaling up from zero nodes. This hasn't been too much of a concern because things usually recover on their own, for example, launching the first user-notebook on the cluster we see (and the user sees these messages while their jupyterhub session is loading). Note in particular `FailedCreatePodSandBox ` followed by `SandboxChanged`:

```
Events:
  Type     Reason                  Age                     From                                    Message
  ----     ------                  ----                    ----                                    -------
  Normal   TriggeredScaleUp        9m26s                   cluster-autoscaler                      pod triggered scale-up: [{eksctl-pangeo-nasa-nodegroup-user-notebook-v2-NodeGroup-8VRX95CW3EOF 0->1 (max: 100)}]
  Warning  FailedScheduling        8m20s (x10 over 9m37s)  nasa-staging-user-scheduler             0/1 nodes are available: 1 node(s) didn't match node selector.
  Warning  FailedScheduling        8m (x7 over 8m18s)      nasa-staging-user-scheduler             0/2 nodes are available: 1 node(s) didn't match node selector, 1 node(s) were not ready.
  Warning  FailedScheduling        7m58s                   nasa-staging-user-scheduler             0/2 nodes are available: 1 node(s) didn't match node selector, 1 node(s) had taints that the pod didn't tolerate.
  Normal   Scheduled               7m58s                   nasa-staging-user-scheduler             Successfully assigned nasa-staging/jupyter-scottyhq to ip-192-168-57-24.ec2.internal
[...]

PR #401, opened 2019-09-08
get icesat2 cluster updates into prod

scottyhq:

PR #402, opened 2019-09-08
Disable continuous-image-puller on AWS hubs

scottyhq:
addresses https://github.com/pangeo-data/pangeo-cloud-federation/issues/398 and https://github.com/pangeo-data/pangeo-cloud-federation/issues/356#issuecomment-529234873

@rsignell-usgs and @jhamman

### DIFF
--- a/deployments/esip/image/binder/environment.yml
+++ b/deployments/esip/image/binder/environment.yml
@@ -29,7 +29,13 @@ dependencies:
   - netcdf4
   - h5py
   - h5netcdf
+  # ocean/atmos stuff
   - siphon
+  - utide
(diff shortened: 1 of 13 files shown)


scottyhq:
@jhamman - I think the commits here are the same as https://github.com/pangeo-data/pangeo-cloud-federation/pull/401 

because I accidentally ran 'Squash and merge' instead of 'Create a merge commit' . Which is why the prod branch currently reports "This branch is 9 commits ahead, 7 commits behind staging." . Just want to confirm that we should be doing "Create a merge commit" and I'll add that to the README so that things stay in sync

Issue #371, opened 2019-08-21
trying out dask-gateway


jhamman:
I got a good way into trying out dask-gateway today. I wanted to share a progress update and rope in a few folks that can help connect a few dots. 

### Deploying

I installed the helm chart onto our main GCP k8s cluster like this:

```bash
helm upgrade --wait --install --namespace=dask-gateway-staging dask-gateway \
    dask-gateway -f dask-gateway-values.yaml
```

where `dask-gateway-values.yaml` is: 
```yaml
gateway:
  proxyToken: "secret"

  clusterManager:
    scheduler:
      cores:
        request: .25
      memory:
        request: 1G

    worker:
      cores:
        request: 2
      memory:
        request: 8G
```

This went smoothly and a few minutes later, I had three pods running in the `dask-gateway-staging` namespace:

```bash
$ kubectl get pods -n dask-gateway-staging
NAME                                            READY   STATUS    RESTARTS   AGE
gateway-dask-gateway-6458b76bfd-6n56w           1/1     Running   1          148m
scheduler-proxy-dask-gateway-6b5f496d84-jm2m5   1/1     Running   0          148m
web-proxy-dask-gateway-5867b9d4dd-5hx6f         1/1     Running   0          148m
$ kubectl get svc --namespace dask-gateway-staging
NAME                            TYPE           CLUSTER-IP    EXTERNAL-IP     PORT(S)          AGE
scheduler-api-dask-gateway      ClusterIP      10.4.9.226    <none>          8001/TCP         154m
scheduler-public-dask-gateway   LoadBalancer   10.4.13.122   35.193.80.169   8786:30884/TCP   154m
web-api-dask-gateway            ClusterIP      10.4.5.169    <none>          8001/TCP         154m
web-public-dask-gateway         LoadBalancer   10.4.1.102    34.67.80.148    80:30848/TCP     154m
```

Looking at the `gateway-dask-gateway` pod's logs, we see that things are up and running...

```bash
$ kubectl logs -n dask-gateway-staging gateway-dask-gateway-6458b76bfd-6n56w -f
[I 2019-08-20 21:28:14.968 DaskGateway] Generating new cookie secret
[I 2019-08-20 21:28:14.970 DaskGateway] Connecting to dask gateway scheduler proxy at 'tls://10.4.13.122:8786', api at 'http://10.4.9.226:8001'
[...]

PR #409, opened 2019-09-20
[WIP] Add dask-gateway configuration

jcrist:
This adds a dask-gateway deployment to the dev hub. This required a
small change to the JupyterHub configuration, as well as the addition of
the dask-gateway helm chart. Things are still rough. TODO:

- [ ] Integrate as a subchart, moving gateway config into normal `common.yaml` file.
- [x] Update images to include new dask-gateway release
- [ ] Add default user-side `~/.config/dask/gateway.yaml` file to image to provide a nicer user configuration (e.g. no need to enter addresses manually).


jcrist:
For the most part this just followed the instructions written here: https://gateway.dask.org/install-kube.html. In particular, the changes to the jupyterhub config were to enable JupyterHub authentication (see https://gateway.dask.org/install-kube.html#authenticating-with-jupyterhub).

dask-values.yaml
```diff
@@ -0,0 +1,42 @@
+gateway:
+  proxyToken: <SECRET>
```
jcrist, line 2: When this config is put in the right spot (rather than a loose file in the top directory), both `<SECRET>` values will be in the `secrets` folder, this was just to get something up for people to see.

jhamman:
I'm going to merge this so we stop breaking the gateway proxy. I'll add the remaining two todo items to the #371.

TomAugspurger:
Do we need to do a few more things to get the deployment to stick? Like `helm install ... dask-gateway/dask-gateway` as part of the deployment? I'm not sure what all gets changed between deployments.

jhamman:
the install of dask-gateway should be stable although it wont be updated as part of the CICD system in this repo.

PR #414, opened 2019-09-24
update Dockerfile to pull image with dask 2.4

rsignell-usgs:


rabernat:
FYI, if you use a newer base image, you get Dask 2.4 with that. See https://github.com/pangeo-data/pangeo-stacks/pull/80

PR #417, opened 2019-09-27
Add GPU node pool for hydro.pangeo.io

jhamman:
This adds the basic config to (hopefully) allow us to use GPUs in our Jupyter sessions. 

I mostly followed https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/994 and https://zero-to-jupyterhub.readthedocs.io/en/latest/user-resources.html?highlight=GPU#set-user-gpu-guarantees-limits (thanks @consideRatio and @yuvipanda!)

With the help of @jsadler2, we now have a alpha docker image with GPU libs in it: https://github.com/pangeo-data/pangeo-stacks/pull/83#issuecomment-535732363

I've added a simple GPU node pool to our cluster, with the following specs:

```
gcloud container node-pools describe --cluster  dev-pangeo-io-cluster  jupyter-gpu-pool                                      ✔  10093  20:59:13
autoscaling:
  enabled: true
  maxNodeCount: 1
config:
  accelerators:
  - acceleratorCount: '1'
    acceleratorType: nvidia-tesla-v100
  diskSizeGb: 100
  diskType: pd-standard
  imageType: COS
  labels:
    hub.jupyter.org/node-purpose: user
  machineType: n1-standard-4
  metadata:
    disable-legacy-endpoints: 'true'
  oauthScopes:
  - https://www.googleapis.com/auth/devstorage.read_only
  - https://www.googleapis.com/auth/logging.write
  - https://www.googleapis.com/auth/monitoring
  - https://www.googleapis.com/auth/servicecontrol
  - https://www.googleapis.com/auth/service.management.readonly
  - https://www.googleapis.com/auth/trace.append
  preemptible: true
  serviceAccount: default
  taints:
  - effect: NO_SCHEDULE
    key: nvidia.com/gpu
    value: 'true'
  - effect: NO_SCHEDULE
    key: hub.jupyter.org_dedicated
    value: user
  - effect: NO_SCHEDULE
    key: nvidia.com/gpu
    value: present
instanceGroupUrls:
- https://www.googleapis.com/compute/v1/projects/pangeo-181919/zones/us-central1-b/instanceGroupManagers/gke-dev-pangeo-io-cl-jupyter-gpu-pool-7f5f1975-grp
management:
  autoRepair: true
  autoUpgrade: true
maxPodsConstraint:
  maxPodsPerNode: '110'
name: jupyter-gpu-pool
podIpv4CidrSize: 24
[...]

PR #415, opened 2019-09-26
bump pangeo chart version

jhamman:

PR #422, opened 2019-10-01
#404, #410, #411, #412, #413, #409, #415, #417, #419, #420, #421 --> prod

jhamman:
This is a big update.


jhamman:
I'm going to merge this now. I've tested all the GCP hubs and things seem happy. I'm also testing the AWS hubs since I'm hypothesizing that just their staging deployments are broken.

PR #439, opened 2019-10-08
Limit number of worker pods per Dask cluster

TomAugspurger:
Users have been crashing ocean.pangeo.io by requesting a large number
of workers. This will limit users to 100 workers.

It's quite easy to work around; a longer-term solution is to use
dask-gateway.

---

If a user does `cluster.scale(400)`, a warning should be printed to the screen saying they requested too many workers.

### DIFF
--- a/deployments/ocean/image/binder/dask_config.yaml
+++ b/deployments/ocean/image/binder/dask_config.yaml
@@ -10,6 +10,8 @@ distributed:
       limit: 5s
 
 kubernetes:
+  count:
+    max: 100
   worker-template:
     metadata:
     spec:
--- a/deployments/ocean/image/binder/dask_config.yaml
+++ b/deployments/ocean/image/binder/dask_config.yaml
@@ -11,7 +11,7 @@ distributed:
 
 kubernetes:
   count:
-    max: 100
+    max: 30
   worker-template:
     metadata:
     spec:


TomAugspurger:
cc @rabernat, this is just a stopgap. I can also push on dask-gateway things soon which will provide a proper fix.

Are there any other deployments this config value should be added to?

rabernat:
Thanks a lot @TomAugspurger! This is an excellent stopgap measure. But let's cap it at 30 workers per cluster.

rabernat:
Can we also add the config that gives a meaningful name to the dask cluster and workers, so we can identify who is doing this? 

Fix by @jhamman described here:
https://github.com/dask/dask-kubernetes/issues/173#issuecomment-527208045

But I'm not sure what section of the config file that should go in.

TomAugspurger:
I *think* we may be doing that

```
>>> cluster = KubeCluster(n_workers=1)
>>> cluster.workers[0].spec.containers[0].name
'dask-0000-0002-8136-7087'
```

That `0000-0002-8136-7087` is the `$JUPYTERHUB_USER` environment variable, which I think comes from `JUPYTERHUB_CLIENT_ID`. Can that be tied back to my ORCID login in the jupyterhub database? I'd look myself, but I don't think I'm an admin on ocean.pangeo.io.

---

@jhamman how do you want to coordinate between this and dask-gateway? As I said on the call I'm free to pick up your dask-gateway stuff starting tomorrow afternoon or Thursday. Should we merge this in the meantime?

TomAugspurger:
As a non-scientific test, I restarted my hub server and I had the same numeric ID `0000-0002-8136-7087`. So hopefully that's stable.

rabernat:
Yes, 0000-0002-8136-7087 is the user-id. If I have that, it's enough.

jhamman:
> As I said on the call I'm free to pick up your dask-gateway stuff starting tomorrow afternoon or Thursday.

Let's just merge this. I think we're a ways out from turning off the daskkubernetes rbac and soley using dask-gateway. This is a nice short term fix.

rabernat:
Thanks Joe! But do you know if we can also fix the user-id thing, as you proposed in https://github.com/dask/dask-kubernetes/issues/173#issuecomment-527208045? It seems like a quick and easy way to be able to debug dask problems more easily.

jhamman:
@rabernat - live in prod as of https://github.com/pangeo-data/pangeo-cloud-federation/pull/441

PR #440, opened 2019-10-08
set dask-worker-name in ocean/dev hubs

jhamman:


jhamman:
cc @rabernat and @TomAugspurger

PR #447, opened 2019-10-14
Increase start_timeout on ocean

jhamman:
(also synced with prod)


rabernat:
FYI we are getting timeouts today even with the 10 minute startup time.

PR #458, opened 2019-10-18
Use released version of hubploy from PyPI

yuvipanda:
Brings in https://github.com/yuvipanda/hubploy/pull/28. Should
be fully backwards compatible!


jhamman:
Thanks @yuvipanda!

PR #457, opened 2019-10-17
Add nbdime to binder environment.yml.

brian-rose:
Not sure if this will install the Jupyterhub / Jupyterlab extension automatically or not.

Potentially closes #456

PR #461, opened 2019-10-18
Add jupyter install step to postbuild

brian-rose:
This will hopefully install the nbdime extension to jupyterlab during the build and avoid prompting the user about it.

Fixes #457


brian-rose:
I checked staging.ocean.pangeo.io and it looks good! The nbdime buttons show up without any prompting.

I think we're good to go.

PR #463, opened 2019-10-18
staging -> prod

tjcrone:
I'm ready to have my changes merged into prod, but I will not merge this until @rabernat and @brian-rose confirm that the changes they made are ready.

I wonder if it would be good to merge into prod as soon as possible and revert if this doesn't happen within a certain amount of time? @yuvipanda and I were talking in Seattle about ways to automate this process. Might be worth looking into.


brian-rose:
As far as I can tell, my changes are working fine.

PR #464, opened 2019-10-19
Send individual notification when pipeline fails

tjcrone:
This adds the ability to send an email from inside the CircleCI configuration, and sets up the CircleCI configuration to email me when the OOI deployment pipeline fails. As suggested [here](https://support.circleci.com/hc/en-us/articles/360007444314-Sending-email-from-a-container), this uses SendGrid to send email. We will need to add the Pangeo SendGrid API key to the CircleCI environment variables to make this work.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -266,9 +266,12 @@ jobs:
       - run:
           name: Deploy ooi.pangeo.io
           when: always
-          command: |
           no_output_timeout: 1200
-            errormsg() { ./.circleci/send-email "tjcrone@gmail.com" "Deployment to ooi-${CIRCLE_BRANCH} failed"; }
+          command: |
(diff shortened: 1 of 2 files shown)


tjcrone:
Just refactored the email script so that it is more general, and accepts an email address, subject line, and email body as arguments. Thoughts? Feelings?

tjcrone:
I am interested in getting input on this. Any thoughts @jhamman? It's not clear to me why this is failing, as it should do nothing as long as the OOI deployment does not fail. It appears that the failure might be inherited from my fork, where I was forcing failure to test the script, and that seems weird but it's possible. I think this can be merged with no impact but I am interested to hear thoughts on the approach. This script might be useful in other contexts as well. Thanks!

PR #471, opened 2019-11-16
staging -> prod (#468,#469,#470)

jhamman:
cc @jbusecke, @rabernat


tjcrone:
Looks good to me. Can we merge?

jbusecke:
There seems to be an issue with some [IPSL outputs](https://github.com/NCAR/intake-esm/issues/181), which I cant reproduce on ocean.pangeo.io, but given the advantage of having `preprocess` available in `intake_esm`, I think that we can go and merge.

PR #472, opened 2019-11-19
update dask-gateway to 0.5.0

TomAugspurger:


TomAugspurger:
@jcrist do you recall why you had the `dask-values.yaml` file split out separately before?

edit: was it because you manually `helm install`ed dask-gateway, rather than through CI?

jcrist:
Yeah, I didn't know how to integrate it with the deployment infrastructure without possibly breaking things.

TomAugspurger:
FYI, still one more issue to sort out here before this is ready. Right now the versions of dask and compression libraries like blosc, lz4 are easy to get out of sync between the client (coming via pangeo-stacks) and the cluster (coming via dask-gateway).

Still thinking through the best way to handle this.

jcrist:
Why not use the same image everywhere? This is what is currently done on pangeo with dask-kubernetes, and seems to work fine. I posted a snippet somewhere of how to configure dask-gateway to automatically use the current client image for the cluster, let me see if I can find it.

TomAugspurger:
Ahh... I may have confused myself with all the images. It doesn't really matter what version of dask-gateway is used for the server, etc. The only one that matters is `gateway.clusterManager.image`.

TomAugspurger:
OK, things are fine w.r.t. the image. We can just reuse the pangeo-stacks image.

The last thing I don't know about is how to test to integrate this with the CI setup here. I've been experimenting locally by running

```
        helm upgrade --install \
                --namespace dask-gateway \
                --version 0.5.0 \
                --values deployments/dev/config/staging.yaml --values deployments/dev/secrets/staging.yaml \
                dask-gateway \
                dask-gateway/dask-gateway

```

Is this something that should be done via hubploy? Or should I just add the helm commands to the `.circleci/config` after the `hubploy`? cc @yuvipanda if you have thoughts on that.

jhamman:
@TomAugspurger - thanks for keeping an eye one this. I'd like to suggest refocusing effort on to https://github.com/pangeo-data/helm-chart/pull/102. Once that PR is in place, dask-gateway will be seamlessly deployed here (with some modifications to the deployment values). Thoughts?

yuvipanda:
+1 to @jhamman

TomAugspurger:
Closing in favor of https://github.com/pangeo-data/helm-chart/pull/104

PR #477, opened 2019-11-26
dev gateway config

jhamman:
This adds dask-gateway to the dev deployment. Now that dask-gateway is part of the pangeo chart, all we need to do is to add the config below to the secrets section of each hub:

```yaml
pangeo:
  jupyterhub:
    ...
    hub:
      services:
        dask-gateway:
          apiToken: SECRET1
  dask-gateway:
    gateway:
      proxyToken: "SECRET2"
      auth:
        type: jupyterhub
        jupyterhub:
          apiToken: "SECRET1"
```

 cc @jcrist, @TomAugspurger


TomAugspurger (approved):
I think you could also remove https://github.com/pangeo-data/pangeo-cloud-federation/blob/staging/dask-values.yaml

TomAugspurger:
Do you also need to add a `helm update` for dask-gateway in https://github.com/pangeo-data/pangeo-cloud-federation/blob/staging/.circleci/config.yml#L228? Or do we get it indirectly via pangeo?

scottyhq:
Nice! 

Would be great to follow up with a comment here demonstrating how to use the new functionality https://github.com/pangeo-data/pangeo-cloud-federation/issues/371. 

I haven't been following closely, but it seems like dask-labextension probably doesn't work with gateway yet (https://github.com/dask/dask-gateway/issues/55#issuecomment-539562108). 

Can we set configuration values here: https://github.com/pangeo-data/pangeo-stacks/blob/master/base-notebook/binder/dask_config.yaml ?

jcrist:
> I haven't been following closely, but it seems like dask-labextension probably doesn't work with gateway yet (dask/dask-gateway#55 (comment)).

It works (kind of) with the master branch of dask-labextension, but not with the latest release. The fix was needed in dask-labextension not in dask-gateway.

> Can we set configuration values here: https://github.com/pangeo-data/pangeo-stacks/blob/master/base-notebook/binder/dask_config.yaml ?

Yes, dask-gateway supports configuration settings the same way as other dask projects. See https://gateway.dask.org/configuration-user.html for more information.

At a minimum, you'll probably want to fix configuration to set the addresses (`gateway.address`, `gateway.proxy-address`) as well as set `gateway.auth.type` to `jupyterhub`. This will allow users to create a gateway cluster with no parameters:

```
from dask_gateway import GatewayCluster

cluster = GatewayCluster()
```

jcrist:
> Would be great to follow up with a comment here demonstrating how to use the new functionality.

For overall usage information, see https://gateway.dask.org/usage.html. Docs are still being written, so if you have questions that aren't answered here please let me know so we can improve them.

jhamman:
Merging to staging. We'll work on getting the gateway config setup once things are deployed there.

PR #487, opened 2019-12-05
prod <- staging

tjcrone:
I'm ready to merge these when you are. My production deployment will likely need to be deleted for this to work, as was required for staging. So when you are ready to merge, please let me know, and I will delete ooi-prod and merge. Thanks!


requirements.txt
```diff
@@ -1,4 +1,4 @@
-git+https://github.com/tjcrone/hubploy@helm-options
+git+https://github.com/yuvipanda/hubploy
```
tjcrone, line 1: @yuvipanda, any reason why we should not use hubploy@master here?

tjcrone:
Apparently we have users who are using ooi-prod to make their AGU posters today and tomorrow. I am now leaning toward merging this after AGU. @jhamman, what are your thoughts on this? Are you considering merging staging into prod before AGU? Thanks.

jhamman:
@tjcrone - I'd like to wait until #481 is done. At a minimum, we need to get the dask-gateway secrets setup for the prod deployments before merging this (or things will break).

tjcrone:
Ah right. The secrets. Based on your input and the fact that breaking prod a few days before AGU would be bad, I am going to close this PR. We can work on this merge after AGU if that works for you.

PR #492, opened 2020-01-09
increase ocean build timeouts

rabernat:
This will trigger a new image to be built for ocean.


rabernat:
Closing in favor of #493.

Issue #500, opened 2020-01-22
Deployments failing


TomAugspurger:
dev, ocean, and hydro are failing on staging.
OOI is failing on prod

Staging: https://circleci.com/gh/pangeo-data/pangeo-cloud-federation/1010
Prod: https://circleci.com/gh/pangeo-data/pangeo-cloud-federation/1001

```
UPGRADE FAILED
Error: Deployment.apps "gateway-dev-staging-dask-gateway" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"app.kubernetes.io/component":"gateway", "app.kubernetes.io/instance":"dev-staging", "app.kubernetes.io/managed-by":"Tiller", "app.kubernetes.io/name":"dask-gateway"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable && Deployment.apps "scheduler-proxy-dev-staging-dask-gateway" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"app.kubernetes.io/component":"scheduler-proxy", "app.kubernetes.io/instance":"dev-staging", "app.kubernetes.io/managed-by":"Tiller", "app.kubernetes.io/name":"dask-gateway"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable && Deployment.apps "web-proxy-dev-staging-dask-gateway" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"app.kubernetes.io/component":"web-proxy", "app.kubernetes.io/instance":"dev-staging", "app.kubernetes.io/managed-by":"Tiller", "app.kubernetes.io/name":"dask-gateway"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable
[...]

PR #501, opened 2020-01-22
bump image version

TomAugspurger:


TomAugspurger:
cc @jhamman this should fix #500, at least for staging.

I think we'll need to manually delete the dask-gateway deployments on prod before merging staging into prod.

jhamman:
Thanks @TomAugspurger! This bump worked on the redeploy. Can you repeat the steps for prod?

TomAugspurger:
@jhamman it looks like prod is on dask-gateway 0.6.0, so that shouldn't be necessary. I'll keep an eye out for a similar failure next time prod is deployed though.

jhamman:
Thanks. I'll setup a prod bump now.

PR #502, opened 2020-01-22
staging -> prod (#497, #498, #499, #501)

jhamman:
cc @TomAugspurger


jhamman:
@TomAugspurger, @rabernat, @tjcrone - this passed on prod. We're all healthy for the first time in a while. Thanks Tom for breaking the log jam.

PR #505, opened 2020-01-23
Nest gateway config under dask-gateway key

TomAugspurger:
I don't think this config was being picked up correctly. This is a bit of a shot in the dark at a fix.


TomAugspurger:
Yep, that did it. The `cluster_options()` is being picked up now, and we're getting the right image in our scheduler pod.

PR #506, opened 2020-01-24
Have default dask-gateway image match

TomAugspurger:
This sets the default image for dask-gateway to `JUPYTER_IMAGE_SPEC`, which is probably a good default.


TomAugspurger:
Mmm, this doesn't quite work. The config is executed on the the server, which doesn't have access to the JUPYTER_IMAGE_SPEC. Will think on this a bit more. The `use_local_defaults` seems promising.

PR #507, opened 2020-01-24
Default image 2

TomAugspurger:
Reverts https://github.com/pangeo-data/pangeo-cloud-federation/pull/506 and (hopefully) provides a working alternative.

The goal is to use `JUPYTER_IMAGE_SPEC` as the default image. That env var isn't available on the dask-gateway server, only on the client. So we override the version from the server with a local default in the typical Dask config file (which is on the client and has `JUPYTER_IMAGE_SPEC`).

### DIFF
--- a/deployments/dev/image/binder/dask_config.yaml
+++ b/deployments/dev/image/binder/dask_config.yaml
@@ -49,3 +49,5 @@ gateway:
   proxy-address: tls://35.225.202.35:8786
   auth:
     type: jupyterhub
+  cluster:
+    image: ${JUPYTER_IMAGE_SPEC}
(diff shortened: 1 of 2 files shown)


TomAugspurger:
This also isn't quite right. We end up putting the literal `${JUPYTER_IMAGE_SPEC}` in the dask config, and dask doesn't expand this.

I think what we want is to set the environment variable `DASK_GATEWAY__CLUSTER__OPTIONS__IMAGE = ${JUPYTER_IMAGE_SPEC}` as the container is created, so that the value of that environment variable is actual evaluated image spec.

PR #508, opened 2020-01-24
Set gateway-enforced resource limits

TomAugspurger:
This limits the amount of cores and memory that users can request. Requesting more than that will (I believe) raise an exception.

@jhamman @rabernat do you have suggestions for what reasonable values would be here?

### DIFF
--- a/pangeo-deploy/values.yaml
+++ b/pangeo-deploy/values.yaml
@@ -78,8 +78,10 @@ pangeo:
               handler=option_handler,
           )
 
-        c.UserLimits.max_cores = 100
-        c.UesrLimits.max_memory = "400 GB"
+        userLimts:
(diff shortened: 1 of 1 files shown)


TomAugspurger:
We can also set a maximum number of independent clusters a user can request. Wasn't sure what to do there.

jhamman:
> We can also set a maximum number of independent clusters a user can request. Wasn't sure what to do there.

Let's limit this at 1 for now. We can always increase it when someone shows a use case.

TomAugspurger:
For now I've put these under their own section in the extraConfig. Deployments specifying their own limits would set

```yaml
dask-gateway:
  gateway:
    extraConfig:
      userLimits:
        c.UserLimits... = ...
```

and still inherit the config for the image picker.

PR #520, opened 2020-01-30
Secure gateway with TLS

TomAugspurger:
Thanks to @jcrist for the suggestion. IIUC, we're piggy backing on jupyterhub's TLS.

We'll need to apply the change to the `dask_confg.yaml` to each deployment.

### DIFF
--- a/deployments/dev/image/binder/dask_config.yaml
+++ b/deployments/dev/image/binder/dask_config.yaml
@@ -45,7 +45,7 @@ labextension:
     kwargs: {}
 
 gateway:
-  address: http://34.68.195.134
+  address: https://staging.hub.pangeo.io/services/dask-gateway/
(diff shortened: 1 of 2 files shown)


jcrist:
This serves the dask-gateway web api through the jupyterhub proxy, which is where jupyterhub terminates TLS. So any web request to dask-gateway goes through `jupyterhub-proxy -> dask-gateway-web-proxy -> backend`. Because these web requests are small in size, I don't expect the load on the jupyterhub proxy to increase significantly.

jcrist:
Note that if still exposed, the old addresss will still work, but will need the prefix added. So `http://34.68.195.134/services/dask-gateway` is still valid. We may want to move this to a `ClusterIP` service to force going through jupyterhub.

PR #537, opened 2020-02-10
staging -> prod (many PRs)

tjcrone:
Most of these PRs updates to the OOI image including a refactor so that it starts with base-notebook rather than pangeo-notebook. We heard from @cspencerjones previously regarding xlayers working in staging. @rabernat, does this cover fastjmd95?


tjcrone:
I think this should be good to go. I will merge early in the morning unless I hear otherwise. Thanks!

PR #562, opened 2020-03-12
update https/proxy configs

jhamman:
This PR updates our proxy/service configs to match the current layout in zero-to-jupyterhub: https://zero-to-jupyterhub.readthedocs.io/en/latest/administrator/security.html#set-up-automatic-https

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -208,7 +208,7 @@ jobs:
           name: Install helm
           when: always
           command: |
-            curl https://get.helm.sh/helm-v3.0.3-linux-amd64.tar.gz | \
+            curl https://get.helm.sh/helm-v3.1.2-linux-amd64.tar.gz | \
               tar -xzf -
             sudo mv linux-amd64/helm /usr/local/bin
             helm repo add pangeo https://pangeo-data.github.io/helm-chart/
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -220,19 +220,19 @@ jobs:
           name: Deploy dev.pangeo.io
           when: always
           command: |
-            hubploy deploy dev pangeo-deploy ${CIRCLE_BRANCH} --force
+            hubploy deploy dev pangeo-deploy ${CIRCLE_BRANCH} --force --atomic --cleanup-on-fail
 
       - run:
           name: Deploy ocean.pangeo.io
           when: always
           command: |
-            hubploy deploy ocean pangeo-deploy ${CIRCLE_BRANCH} --force
+            hubploy deploy ocean pangeo-deploy ${CIRCLE_BRANCH} --force --atomic --cleanup-on-fail
 
       - run:
           name: Deploy hydro.pangeo.io
           when: always
           command: |
-            hubploy deploy hydro pangeo-deploy ${CIRCLE_BRANCH} --force
+            hubploy deploy hydro pangeo-deploy ${CIRCLE_BRANCH} --force --atomic --cleanup-on-fail
 
       # - run:
       #     name: Deploy nasa.pangeo.io
@@ -255,7 +255,7 @@ jobs:
           name: Deploy aws-uswest2.pangeo.io
           when: always
           command: |
-            hubploy deploy icesat2 pangeo-deploy ${CIRCLE_BRANCH} --force
+            hubploy deploy icesat2 pangeo-deploy ${CIRCLE_BRANCH} --force --atomic --cleanup-on-fail
       - run:
           name: Revert to Original EKS IP Whitelist
           when: always
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -220,19 +220,19 @@ jobs:
           name: Deploy dev.pangeo.io
           when: always
           command: |
-            hubploy deploy dev pangeo-deploy ${CIRCLE_BRANCH} --force --atomic --cleanup-on-fail
+            hubploy deploy dev pangeo-deploy ${CIRCLE_BRANCH} --force --atomic
 
       - run:
           name: Deploy ocean.pangeo.io
           when: always
           command: |
-            hubploy deploy ocean pangeo-deploy ${CIRCLE_BRANCH} --force --atomic --cleanup-on-fail
+            hubploy deploy ocean pangeo-deploy ${CIRCLE_BRANCH} --force --atomic
 
       - run:
           name: Deploy hydro.pangeo.io
           when: always
           command: |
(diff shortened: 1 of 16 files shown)


tjcrone (approved):

tjcrone:
These all look good to me. We are in a good spot for testing all of this on OOI. Happy to assist as needed. Thanks @jhamman!

consideRatio:
I think the service to be configured to get an external IP and act as the entrypoint of traffic, need to be the proxy-public service, but this PR makes the hub service become the external entrypoint.

The hub service will go to the hub pod, and the hub pod wont proxy traffic onwards. The hub pod does not understand how to respond to someone requesting /user/erik. The hub pod do know how to accept a user on /hub/home and later on demand configure the proxy pod (which gets traffic from proxy public) to route /user/erik properly to the individual user pod. But, for the proxy pod to route to /user/erik, it needs to get the incoming traffic redirect to it, but that requires it to be the entrypoint of the domain, but this PR makes the Hub get the traffic to itself without passing through the proxy.

I didn't see a reference to this configuration in the documentation.

Hmmm... But with autohttps, there is one more proxy that shows up in front of the proxy-public service i think, a pod that will decrypt/encrypt incoming/outgoing traffic. So, the key here to ensure incoming traffic goes there i guess, but i think that should not force the user of the chart make a change like this.

scottyhq:
yeah @jhamman and @consideRatio  i was just wondering why this change would be needed given how things have been working in the past?... 

I find it confusing that `loadBalancerIP` can be under either proxy or hub, but the documentation doesn't really clarify how they are different:
https://zero-to-jupyterhub.readthedocs.io/en/stable/reference.html#proxy-service-loadbalancerip
https://zero-to-jupyterhub.readthedocs.io/en/stable/reference.html#hub-service-loadbalancerip

consideRatio:
@scottyhq yes I think that has been added for completeness at some point as the proxy service had that option, but it doesn't make sense to me in a z2jh deployment that one would ever configure the hub service to be a external entrypoint for network traffic.

jhamman:
@consideRatio - to be honest, I'm not following all of what you and @scottyhq are going over. Are you suggesting I move/remove the `loadBalancerIP` section from our config?

requirements.txt
```diff
@@ -1,4 +1,4 @@
-git+https://github.com/yuvipanda/hubploy@a02fb01bbc4a698b2cc85f9677e9881dac8d438a
+git+https://github.com/jhamman/hubploy@c3f270f11c4aa60314a125fdd1ebbed65b8a643b
```
jhamman, line 1: see also: https://github.com/yuvipanda/hubploy/pull/56

deployments/dev/config/staging.yaml
```diff
@@ -3,12 +3,12 @@ pangeo:
...
       https:
         hosts:
-        - staging.hub.pangeo.io
+          - staging.hub.pangeo.io
         letsencrypt:
           contactEmail: jhamman@ucar.edu
+    hub:
       service:
         loadBalancerIP: 35.232.186.70
-    hub:
```
consideRatio: @jhamman I mean that this moving of the proxy.service.loadBalancerIP to hub.service.loadBlanacerIP will break things. The hub k8s service should not have the external IP of the deployment.

jhamman: Gotcha. So should I just remove the `loadBalancerIP` all together? Is this needed anymore?

consideRatio: I think keep it and don't change any staging.yaml or prod.yaml.

> This PR updates our proxy/service configs to match the current layout in zero-to-jupyterhub: https://zero-to-jupyterhub.readthedocs.io/en/latest/administrator/security.html#set-up-automatic-https

I've not understood what you referred to so I'm very confused in general - what is the purpose of this PR? As far as I know, there shouldn't be any change to the proxy.service.loadBalancerIP configuration due to an upgrade from 0.9.0-beta.2 to 0.9.0-beta.3 or 4, and I'm confident that the proxy.service.loadBalancerIP configuration should not be moved to hub.service.loadBalancerIP.

jhamman: the confusion on my end is that `loadBalancerIP` is not mentioned anywhere on the page linked above. Based on those docs, its not clear to me if `loadBalancerIP` is needed or where it should be applied.

consideRatio: Ah I see!

It is used when a service is of `type: LoadBalancer`, then you can request a specific IP if that IP is available for use. That way you ensure you always have the same IP for the various deployments, which will be very relevant for domain names referencing fixed IPs.

So, to summarize, it is not essential for HTTPS, but it is a recommended practice to ensure you don't end up with a different IP if you reinstall this Helm chart or similar. I think it would make sense to have documentation or a mention about this in https://zero-to-jupyterhub.readthedocs.io/en/latest/administrator/security.html#set-up-automatic-https!

jhamman: Got it. I've rolled back all the loadBalancerIP changes.

jhamman: And yes! it would be great to tie these two pieces of documentation together.

deployments/ocean/config/staging.yaml
```diff
@@ -3,12 +3,12 @@ pangeo:
...
       https:
         hosts:
-        - staging.ocean.pangeo.io
+          - staging.ocean.pangeo.io
         letsencrypt:
           contactEmail: raphael.dussin@gmail.com
+    hub:
       service:
         loadBalancerIP: 35.238.36.78
-    hub:
```
consideRatio: This also

.circleci/config.yml
```diff
@@ -255,7 +255,7 @@ jobs:
           name: Deploy aws-uswest2.pangeo.io
           when: always
           command: |
-            hubploy deploy icesat2 pangeo-deploy ${CIRCLE_BRANCH} --force
+            hubploy deploy icesat2 pangeo-deploy ${CIRCLE_BRANCH} --force --atomic --cleanup-on-fail
```
consideRatio: I dont think using both --atomic and --cleanup-on-fail will make sense, atomic is doing what cleanup does but a bit more than the cleanup flag as i understand it.

I suggest using only the atomic flag to have more comprehensible code down the line.

deployments/dev/config/staging.yaml
consideRatio: I opened this issue to represent that: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1599

PR #568, opened 2020-03-18
Changes to move us forward on dask-gateway

tjcrone:


tjcrone:
Should we come up with a naming scheme for DNS entries for the Pangeo dask-gateway proxies? Possibly `dask.ooi.pangeo.io` and `dask.staging.ooi.pangeo.io`. Or something similar.

deployments/ooi/config/common.yaml
```diff
@@ -37,6 +37,10 @@ pangeo:
         guarantee: 1.25G
     hub:
       #nodeSelector: {agentpool: nodepool1}
+      services:
+        dask-gateway:
+          # This makes the gateway available at ${HUB_URL}/services/dask-gateway
+          url: "http://web-public-ooi-staging-dask-gateway"
```
scottyhq: You'll want "http://web-public-ooi-prod-dask-gateway" as well right? So this should go into both prod.yaml and staging.yaml instead of common.yaml?

deployments/ooi/config/staging.yaml
```diff
@@ -13,3 +13,10 @@ pangeo:
       service:
         annotations: {'service.beta.kubernetes.io/azure-load-balancer-resource-group': 'ooi-pangeo'}
         loadBalancerIP: 13.92.63.24
+  dask-gateway:
+    webProxy:
+      service:
+        loadBalancerIP: 13.82.106.226
+    schedulerProxy:
+      service:
+        loadBalancerIP: 52.168.133.15
```
scottyhq: Am I understanding correctly that labeling these and creating DNS entries is only necessary for bypassing JupyterHub https? Or is this necessary to simplify the user-facing dashboard addresses?

scottyhq (commented):
Thanks for moving this forward @tjcrone !

> Should we come up with a naming scheme for DNS entries for the Pangeo dask-gateway proxies? Possibly dask.ooi.pangeo.io and dask.staging.ooi.pangeo.io. Or something similar.

Seems reasonable to me. How will people interact directly with these addresses?

deployments/ooi/config/staging.yaml
scottyhq: As someone not so familiar with networking I'm always unsure what IP addresses should be public or not... so as a precaution (in case authentication and https isn't properly set up) I'm leaning towards hiding these to the extent possible. So how about putting loadBalancerIPs under secrets/staging.yaml instead of config/staging.yaml ?

deployments/ooi/config/common.yaml
tjcrone: Ah right good catch. This needs to be split. Thanks!

deployments/ooi/config/staging.yaml
tjcrone: Well, as far as I know, users will need the proxy IP address to create a gateway, and having a DNS entry might provide a better user experience:
```
gateway = Gateway(address='https://staging.ooi.pangeo.io/services/dask-gateway',
                  proxy_address='tls://52.168.133.15:8786',
                  auth='jupyterhub')
```
I don't know if this can be obscured. But maybe. Also the best user experience is with the extension, so maybe there is a way to automatically populate the extension. Regarding secrets, I don't see any reason why these IP addresses would need to be secret. @jcrist do you have any thoughts/guidance on this? Thanks!

jcrist: > Regarding secrets, I don't see any reason why these IP addresses would need to be secret.

There's no reason for these to be secret, but I agree that ip addresses aren't the best user experience.

If you're connecting from inside the kubernetes cluster (which is true if you're running on JupyterHub) then you can use the DNS entry that kubernetes created automatically for this service. Something like:

```
proxy_address="tls://scheduler-public-ooi-prod-dask-gateway:8786"
```

To keep things easy for users, you can stash these values in dask's user config (I recommend doing this). This allows users to create a gateway client with no parameters.

```python
import dask_gateway

gateway = dask_gateway.Gateway()
cluster = gateway.new_cluster()

# Or create a dask-gateway cluster, skipping using the gateway client entirely
cluster = dask_gateway.GatewayCluster()
```

The dev instance configures this here: https://github.com/pangeo-data/pangeo-cloud-federation/blob/e55e623ba9bcbd506f3a5c40e65f69a9a2689fd3/deployments/dev/image/binder/dask_config.yaml#L47-L54. See https://gateway.dask.org/configuration-user.html#default-configuration for docs on all user-side configuration options.

scottyhq: @jcrist - thanks for the additional info! One issue with putting the config into image/binder/dask_config.yaml is that many of the hubs are now using multiple images in a profile_list. It would be best to set these URLs for any image on the hub, so maybe we copy dask_config.yaml under initContainers? https://github.com/pangeo-data/pangeo-cloud-federation/blob/staging/deployments/icesat2/config/common.yaml#L16

jcrist: Could also set it via environment variable - any dask configuration value can be set by a corresponding environment variable (see https://docs.dask.org/en/latest/configuration.html). For example:

```
DASK_GATEWAY__ADDRESS="https://staging.hub.pangeo.io/services/dask-gateway/"
```

These could be added via the pod spec, so they'd apply to any image.

TomAugspurger: Just FYI, https://zero-to-jupyterhub.readthedocs.io/en/latest/customizing/user-environment.html#set-environment-variables may be helpful here. I'm trying that out on the binder deployment now (https://github.com/pangeo-data/pangeo-binder/pull/110/files)

TomAugspurger: (obviously, valid syntax helps: https://github.com/pangeo-data/pangeo-binder/pull/111 😄)

tjcrone:
Thank you @jcrist! This is super helpful information. Since I most if not all users should be connecting from inside the kubernetes cluster, and because `proxy_address="tls://scheduler-public-ooi-prod-dask-gateway:8786"` is an option, I am going to remove the static IP addresses, and drop the idea of assigning DNS entries at this stage. So this solves a bunch of issues all at once. Great!

I think the next things we should think about are figuring out how to get the dask labextension to integrate more seamlessly with dask-gateway, and maybe having Client report the correct dashboard URL. But this is big progress. Thank you!

PR #572, opened 2020-03-25
Cleanup dev-staging config

TomAugspurger:
* Moves config to environment variables rather than dask_config.yaml
* Organizes a few common config values that apply everywhere, and a few deploy-specific config values

### DIFF
--- a/deployments/dev/image/binder/dask_config.yaml
+++ b/deployments/dev/image/binder/dask_config.yaml
@@ -43,12 +43,3 @@ labextension:
     class: KubeCluster
     args: []
     kwargs: {}
-
-gateway:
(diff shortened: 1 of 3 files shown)


deployments/dev/image/binder/dask_config.yaml
```diff
@@ -43,12 +43,3 @@ labextension:
     class: KubeCluster
     args: []
     kwargs: {}
-
-gateway:
-  address: https://staging.hub.pangeo.io/services/dask-gateway/
-  proxy-address: tls://35.225.202.35:8786
-  auth:
```
TomAugspurger, line 50: `auth` was incorrectly dropped here. Fixing now.

PR #574, opened 2020-03-25
ocean gateway

TomAugspurger:


TomAugspurger:
CI is currently timing out on the conda environment solving. @scottyhq how close to being ready is pangeo-stacks-dev? :)

For now, I'll try to help out the solver...

scottyhq:
> CI is currently timing out on the conda environment solving. @scottyhq how close to being ready is pangeo-stacks-dev? :)

I'll work on it today. There are some changes I'd like to make, but the existing images are functional as far as I know. So if you want you could use a config like this https://github.com/scottyhq/pangeodev-binder/tree/master/binder 

I think we should start pushing images to pangeo/ instead of pangeodev/ though...

TomAugspurger:
https://github.com/pangeo-data/pangeo-cloud-federation/pull/574/commits/784848ea575abcccb362352ce2fee8c182f5e52a has the fully-resolved environment output. I  generated it by pulling pangeo-notebook locally and running the `conda env update -p environment.yaml` in a docker image, then `conda env export --no-build` on the result.

I'm not sure that we actually want to include the fully specified environment, but I don't know how else to actually get the conda solve to finish (if it even does).

TomAugspurger:
OK that helped conda. The solve finished in ~11 minutes (vs. 1.5 hours).

I think the next thing to do is find which packages are causing problems.

As an aside, dfid people know that tensorflow was being pulled in?

TomAugspurger:
I'm going to go ahead with this now. For reference, https://github.com/pangeo-data/pangeo-cloud-federation/pull/574/commits/0115869e1041b1c11d1697e210176048665eb41b pins packages with >200 dependents (checking the output of `conda env update -p notebook -f environment.yaml -vv` for lines like `DEBUG conda.resolve:filter_group(538): pandas: pruned from 228 -> 29`). I don't think any versions have changed aside from adding dask-gateway. Just pinned the versions that were solved for.

Those pins to help the conda solver may come back to bite us when we bump the version of pangeo-notebook in `ocean/image/binder/Dockerfile`. We'll need to update the versions I pinned to match the versions in the updated dockerfile, but hopefully @scottyhq's updates to pangeo-stacks are ready by then.

Issue #584, opened 2020-04-21
Bump to Dask Gateway 0.7.1


TomAugspurger:
Dask-Gateway 0.7.1 made some changes to the helm configuration from 0.6.1. I'd like to get all the hubs moved over to 0.7.1 as a group.

We did this on the binders at https://github.com/pangeo-data/pangeo-binder/pull/139.

IIUC, we'll need to

1. Bump dask-gateway at https://github.com/pangeo-data/helm-chart/blob/6749ba1276f0eacc68bf23be4224504403cd2e99/pangeo/Chart.yaml#L12-L14
2. Depend on the new helm chart at https://github.com/pangeo-data/pangeo-cloud-federation/blob/d927da6594035b5455d565e564418e937ce893e7/pangeo-deploy/Chart.yaml#L8
3. Bump all the dask-gateway in the `environment.yaml`s (e.g. https://github.com/pangeo-data/pangeo-cloud-federation/blob/staging/deployments/dev/image/binder/environment.yml).

cc @scottyhq, @tjcrone.

scottyhq:
@TomAugspurger  testing on staging.aws-uswest2.pangeo.io I'm seeing the following traceback for new_cluster()

```pytb
/srv/conda/envs/notebook/lib/python3.7/site-packages/aiohttp/client_reqrep.py in raise_for_status(self)
    944                 status=self.status,
    945                 message=self.reason,
--> 946                 headers=self.headers)
    947 
    948     def _cleanup_writer(self) -> None:

ClientResponseError: 502, message='Bad Gateway', url=URL('https://staging.aws-uswest2.pangeo.io/services/dask-gateway/api/v1/clusters/')
```

With lots of repeat tracebacks in the logs:
```
[E 2020-04-28 21:38:35.249 KubeController] Error in cluster informer, retrying...
Traceback (most recent call last):
  File "/opt/conda/lib/python3.7/site-packages/dask_gateway_server/backends/kubernetes/utils.py", line 161, in run
    initial = await method(**self.method_kwargs)
  File "/opt/conda/lib/python3.7/site-packages/dask_gateway_server/backends/kubernetes/utils.py", line 59, in func
    return await method(*args, **kwargs)
  File "/opt/conda/lib/python3.7/site-packages/kubernetes_asyncio/client/api_client.py", line 166, in __call_api
    _request_timeout=_request_timeout)
[...]

PR #592, opened 2020-04-28
Set jupyterhub auth

TomAugspurger:
Unsure why this is necessary, given that we set this in the pangeo helm
chart. Perhaps since we override the token in the secret files,
which is a subkey?

### DIFF
--- a/pangeo-deploy/values.yaml
+++ b/pangeo-deploy/values.yaml
@@ -42,6 +42,8 @@ pangeo:
       userPlaceholder:
         enabled: false
   dask-gateway:
+    auth:
+      type: jupyterhub
     gateway:
         # TODO: ClusterManger replacment
         # clusterStartTimeout: 600


pangeo-deploy/values.yaml
```diff
@@ -42,6 +42,8 @@ pangeo:
       userPlaceholder:
         enabled: false
   dask-gateway:
+    auth:
+      type: jupyterhub
```
scottyhq, line 46: should this be under 'gateway'?

scottyhq:
@TomAugspurger I think this deployed correctly but messed up creating clusters https://github.com/pangeo-data/pangeo-cloud-federation/issues/584#issuecomment-621559707

I noticed I've been setting jupyterhub auth as an environment variable, which also seems to work
https://github.com/pangeo-data/pangeo-cloud-federation/blob/2a8f23eb8a9bc06e4d4187a19a32524c0f0dd6f4/deployments/icesat2/config/common.yaml#L19

PR #613, opened 2020-06-09
Use pangeo serviceAccount everywhere

TomAugspurger:
This makes us use the `pangeo` Kubernetes ServiceAccount on

* singleuser pods
* scheduler pods
* worker pods

Previously, we only used it for worker pods, except for icesat which
used it for singleuser & worker.

This is in preparation for granting read / write access to a storage
bucket to the `pangeo` service account on the cloud backends.

cc @scottyhq @tjcrone. I don't think this requires anything on your end, since the `pangeo` Kubernetes Service Account should already exist. Do you see any reason not to use it for users / scheduler pods as well?


TomAugspurger:
Planning to merge this tomorrow. I think we don't want to touch prod the week since the icesat-2 hackweek is running.

scottyhq:
Thanks @TomAugspurger - adding the same serviceaccount to all "user-associated" pods seems good to me. Feel free to merge staging and prod. We're using a separate hub for the hackweek over the next two weeks

PR #614, opened 2020-06-10
draft maintenance announcements

jhamman:
This get's the ball rolling for our announcements related to shutting down hub.pangeo.io, hydro.pangeo.io, and reconfiguring ocean.pangeo.io. 

TODO:
- [x] merge https://github.com/pangeo-data/pangeo-custom-jupyterhub-templates/pull/6
- [x] update our custom configs to point to new custom templates (from PR above)
- [x] settle on announcement text and link to discourse.

cc @rabernat, @TomAugspurger

PR #605, opened 2020-05-12
try generic openid-connect config

rabernat:
After #604 failed, I'm trying a new approach.

Here is `secrets/staging.yaml`.

```
pangeo:
  jupyterhub:
    auth:
      custom:
        config:
          client_id: "XXX"
          client_secret: "XXX"
```

PR #604, opened 2020-05-12
try auth0 on staging.hub

rabernat:
At the cloud ops meeting today, we discussed our scattered approach to auth. After doing some research today, I believe we may be able to get by with the free version of [auth0](https://auth0.com/). I have set up an account and configured an app with the github connector. 

This PR is my attempt to plug it in to staging.hub.pangeo.io.

`secrets/staging.yaml` looks like this:

```yaml
pangeo:
  jupyterhub:
    auth:
      type: auth0
      auth0:
        clientId: "XXX"
        clientSecret: XXX
        callbackUrl: "https://staging.hub.pangeo.io/hub/oauth_callback"
```

If it works, we will immediately hit https://github.com/jupyterhub/oauthenticator/issues/266#issuecomment-551086406: the auth0 authenticator only allows the username to be email. Instead, we want to use `nickname`, which contains the github username. I think that should probably be fixed upstream in oauthenticator.

PR #617, opened 2020-06-10
staging -> prod [#604, #605, #606, #607, #608, #612, #613, #614]

jhamman:
Update prod. Includes: #604, #605, #606, #607, #608, #612, #613, #614


TomAugspurger:
Might want to avoid the auth0 changes? I haven't been able to figure it out yet.

deployments/dev/config/common.yaml
```diff
@@ -68,44 +68,47 @@ pangeo:
...
+        config:
+          login_service: "auth0"
+          token_url: https://pangeo.auth0.com/oauth/token
+          userdata_url: https://pangeo.auth0.com/oauth/userinfo
+          userdata_method: GET
+          username_key: nickname
+          scope:
+            - openid
+            - profile
+            - email
```
jhamman, line 105: as @TomAugspurger pointed out, should probably roll these changes back. @rabernat -- is that something you can do?

rabernat: Happy to help, but I'm not sure exactly what you're asking me to do.

jhamman: I don't think this will work on prod since the callback is to `staging`. So we either need to a) roll this change back or b) split out the config to work for both staging/prod.

TomAugspurger: Will push up a PR to split the config. Hopefully that's all that's needed?

TomAugspurger: That's in. Hopefully it works.

scottyhq: Let me do the same for AWS now before we merge this

scottyhq:
Thanks @jhamman and @TomAugspurger . Happy to have this merged any time

PR #618, opened 2020-06-10
Announcement part 2

jhamman:


rabernat:
Looks like you sorted it out! Thanks Joe!

Let's get these announcements to prod asap.

PR #632, opened 2020-06-27
tolerate preemptible nodes for dask workers

jhamman:
Not deployed yet.


TomAugspurger:
This should already be working now.

PR #648, opened 2020-07-02
Use Prometheus operator for dashboards

TomAugspurger:
This enables metrics using the prometheus-operator helm chart, which wraps prometheus and grafana.

Currently, this isn't deployed via CI. It has to be run locally. We're deploying into a separate namespace, so it's part of a different `helm install` command.

And I still have figured out how to persist helm deploys. I have a `metrics/dashboards/dask.json` file that I'd like to import by default. Maybe @salvis2 has some guesses once we get this running?


deployments/gcp-uscentral1b/config/common.yaml
```diff
@@ -75,6 +80,8 @@ pangeo:
                 }
             },
           ]
+        prometheus: |
+          c.JupyterHub.authenticate_prometheus = False
```
TomAugspurger, line 84: I think this makes it feasible for users within the kubernetes cluster to query the prometheus endpoints. I don't think we care about that, but I can adjust this if desired.

salvis2:
Are we not able / don't want to set `prometheus-operator` as a dependency in Helm? Seems like that would get it into the CI system.

salvis2:
As far as importing dashboards, the grafana Helm Chart (https://github.com/helm/charts/tree/master/stable/grafana#import-dashboards) says that you can set this inside the grafana config:

```
dashboards:
  default:
    custom-dashboard:
      # This is a path to a file inside the dashboards directory inside the chart directory
      file: dashboards/custom-dashboard.json
```

Might be a little trial and error on getting `metrics/dashboards/dask.json` into the right spot, but I think it's almost there.

TomAugspurger:
>Are we not able / don't want to set prometheus-operator as a dependency in Helm? Seems like that would get it into the CI system.

My assumption was that it would fail since hubploy handle things like auth. But I can try it out.

TomAugspurger:
Wasn't able to figure out importing dashboards. I'll have to table this for a bit.

salvis2:
I'm checking out some dashboarding stuff for the ICESat-2 Hackweek cluster. If you are trying to import a dashboard from Grafana's site, you need an additional `dashboardProviders` block: https://github.com/helm/charts/issues/19973#issuecomment-578846218

Edit: I deployed Grafana for the hackweek cluster with this fix and got the desired Grafana dashboard to show up.

salvis2:
I've also found and deployed a custom dashboard's JSON with some info in this issue: https://github.com/helm/charts/issues/12679 

This also depends on the `dashboardProviders` block but the `file` option that grafana claims can work doesn't work with `helm install` (which is mentioned in that issue). You have to paste the entire dashboard JSON into the config file under the `json: |` line, tabbed over (at least once?).

TomAugspurger:
Trying this one more time.

TomAugspurger:
Importing from grafana.com seemed to work. Right now I just have the Dask dashboard. We can add more and upload them to grafana.com.

Earlier I had to make some changes to the queries for the Dask dashboard, I think because the current one isn't designed to handle multiple namespaces.

.circleci/config.yml
```diff
@@ -178,6 +178,7 @@ jobs:
           when: always
           command: |
             hubploy deploy gcp-uscentral1b pangeo-deploy ${CIRCLE_BRANCH} --cleanup-on-fail --namespace ${CIRCLE_BRANCH}
+            helm upgrade --wait --install metrics stable/prometheus-operator --namespace=metrics -f metrics/values.yaml
```
TomAugspurger, line 181: This *might* fail CI. If it does I'll followup with a PR removing it and we'll need to manually update things when the dashboards change.

TomAugspurger:
CI did fail. I've removed the metrics deployment steps.

salvis2:
@TomAugspurger I just did a test and `hubploy` (on the command line) can deploy `prometheus-operator` if we put it as a dependency in `pangeo-deploy/Chart.yaml`. It does deploy into the same namespace as the JupyterHub though. Was that a problem?

TomAugspurger:
Oh nice idea. I'll see if I can incorporate that into https://github.com/pangeo-data/pangeo-cloud-federation/pull/671.

> It does deploy into the same namespace as the JupyterHub though.

There were some namespacing issues, but I don't recall exactly what they were. I'll see what breaks.

We could also just monitor prod and ignore staging.

PR #671, opened 2020-07-17
Add ingress for Grafana dashboard

TomAugspurger:
This makes the grafana dashboard available at http://grafana.us-central1-b.gcp.pangeo.io. ~It requires authentication.~


TomAugspurger:
@salvis2 have you done the metrics for AWS? This has some gcp specific stuff. We'll need to split the `metrics/values.yml` file to have some deployment-specific stuff if / when you do this for AWS.

salvis2:
> @salvis2 have you done the metrics for AWS? This has some gcp specific stuff. We'll need to split the `metrics/values.yml` file to have some deployment-specific stuff if / when you do this for AWS.

Not yet. If we need to have deployment-specific metric config, shouldn't the GCP bits live somewhere like `deployments/gcp-uscentral1b/config/common.yaml`? Note: I haven't had a chance to look closely at how you are setting this up.

TomAugspurger:
Either layout (metrics/deployement vs. deployments/metrics) should work.

salvis2:
One thing I notice on the current dashboard is that it isn't secured. The config I've set up for the binderhubs _is_ secured if you want it. The ingress looks like this:

```
prometheus-operator:
  grafana:
    ingress:
      enabled: true
      hosts:
        - grafana.staging.binder.pangeo.io
      annotations:
        kubernetes.io/ingress.class: nginx
        kubernetes.io/tls-acme: "true"
        cert-manager.io/issuer: letsencrypt-production
      tls:
        - secretName: grafana.staging.binder.binder.io-tls
          hosts:
            - grafana.staging.binder.binder.io
```

I also needed to specify

```
prometheus-operator:
  grafana:
    service:
      type: LoadBalancer
```

To get `cert-manager` to issue certificates for the grafana subdomain. There should be a way to do this without specifying an `ExternalIP` for grafana, but it'll require an IAM role / serviceaccount / some other work, see https://cert-manager.io/docs/configuration/acme/dns01/ . I think this also would like our domain hosting to be somewhere specific and probably not where it is now.

TomAugspurger:
@salvis2 pushed a biggish update in https://github.com/pangeo-data/pangeo-cloud-federation/pull/671/commits/05513849b01bbf10a893a593a1f186e6dc956ce8 to use hubploy by adding prometheus-operator and nginx-ingress as a subchart.

By default we (shouldn't) do anything. You'd need to set `metrics.enabled: true` to bring in the subcharts and enable metrics.

If that's set, along with a couple other variables like https://github.com/pangeo-data/pangeo-cloud-federation/pull/671/files#diff-649bcf98d7669f65efdd8b85fd07298dR32-R47, then things will be enabled.

salvis2:
> By default we (shouldn't) do anything. You'd need to set metrics.enabled: true to bring in the subcharts and enable metrics.

So we need a little logic so that the creation of resources in the `nginx-ingress` and `prometheus-operator` charts are conditional on `metrics.enabled: true`. That sounds not too bad, but is bringing in the charts an easy conditional to set up? A GitHub Action to make the `helm repo add` command optional?

TomAugspurger:
I think the `helm repo add` part is fine to do regardless. We just want to avoid actually installing the `prometheus-operator` subchart (and all it's pods, services etc.) when `metrics.enabled` is false.

So IIUC, that's exactly what the `condition` flag in the Chart's dependencies section controls: https://github.com/pangeo-data/pangeo-cloud-federation/pull/671/files#diff-aef3875c39f29b15ca74987264926466R19

salvis2:
I see. I think I was mashing some thoughts together and thinking those two actions (bring in charts and enabled metrics) would be enabled separately. Makes sense that they should be enabled or disabled together.

TomAugspurger:
@salvis2 can you check my https setup in b9bfbda31ed28fd291cb7e2a0a3c087ce05a7dcb? With that visiting https://grafana.us-central1-b.gcp.pangeo.io says that the certificate is just valid for `ingress.local`, not `grafana.us-central1-b.gcp.pangeo.io`.

Edit: perhaps relevant, from the ingress pod's logs:

```
2020-08-26T19:30:03.181839870Z W0826 19:30:03.181586       6 backend_ssl.go:46] Error obtaining X.509 certificate: no object matching key "staging/grafana.us-central1-b.gcp.pangeo.io.io-tls" in local store
2020-08-26T19:30:04.148754858Z W0826 19:30:04.148488       6 controller.go:916] Service "staging/gcp-uscentral1b-staging-grafana" does not have any active Endpoint.
2020-08-26T19:30:04.148799478Z W0826 19:30:04.148585       6 controller.go:1132] Error getting SSL certificate "staging/grafana.us-central1-b.gcp.pangeo.io.io-tls": local SSL certificate staging/grafana.us-central1-b.gcp.pangeo.io.io-tls was not found. Using default certificate
2020-08-26T19:30:04.148807878Z I0826 19:30:04.148695       6 controller.go:141] Configuration changes detected, backend reload required.
2020-08-26T19:30:04.299129086Z I0826 19:30:04.298899       6 controller.go:157] Backend successfully reloaded.
2020-08-26T19:30:07.482062009Z W0826 19:30:07.481878       6 controller.go:916] Service "staging/gcp-uscentral1b-staging-grafana" does not have any active Endpoint.
2020-08-26T19:30:07.482107360Z W0826 19:30:07.481933       6 controller.go:1132] Error getting SSL certificate "staging/grafana.us-central1-b.gcp.pangeo.io.io-tls": local SSL certificate staging/grafana.us-central1-b.gcp.pangeo.io.io-tls was not found. Using default certificate
```

salvis2:
@TomAugspurger I had a problem enabling HTTPS for my deployment for the binderhub. After some reading, I found I think two solutions. The main question is: What is the solver in the `cert-manager` issuer for this deployment? For reference, here is the [AWS binderhub one](https://github.com/pangeo-data/pangeo-binder/blob/staging/k8s-aws/binderhub-issuer-prod.yaml#L16-L17). I think that using the `http01` solver _needs_ there to be an `EXTERNAL-IP` for Grafana. That can be accomplished with 

```
prometheus-operator:
  grafana:
    service:
      type: LoadBalancer
```

Otherwise, there is some more legwork to set up the [dns01 solver](https://cert-manager.io/docs/configuration/acme/dns01/).

TomAugspurger:
Mmm OK thanks. In that case I might just ignore it for now and come back to it later?

TomAugspurger:
Then we can merge this and actually test it out through hubploy / deploy it to other clusters.

salvis2:
Yeah if it's all working out I don't see why we can't try merging and seeing how it all goes. It was pretty easy for me to get that `EXTERNAL-IP` up and get HTTPS running that way.

I read in the comments that `metrics.enabled` is `False` by default, but where is that set up? Is it just that you have to define `metrics.enabled` at all, since it's just under the umbrella chart that `hubploy` deploys?

TomAugspurger:
> I read in the comments that metrics.enabled is False by default, but where is that set up? Is it just that you have to define metrics.enabled at all, since it's just under the umbrella chart that hubploy deploys?

Yeah. I've done everything in `<deployment>/config/staging.yaml`, just so it's picked up in our next deployments to staging, but it could be anywhere. For AWS I think you'd need something like


```yaml
# file: deployments/icesat2/config/staging.yaml

metrics:
  enabled: true

prometheus-operator:
  grafana:
    server:
      domain: "grafana.aws-uswest2.pangeo.io/"
    ingress:
      hosts:
        - "grafana.staging.aws-uswest2.pangeo.io/"

ingress-nginx:
  controller:
    service:
      loadBalancerIP: "<TODO>" 
```

and

```yaml
# file: deployments/icesat2/secrets/staging.yaml
prometheus-operator:
  grafana:
    adminPassword: "<token>"
```

I think that's it.

salvis2:
Looks like it deployed, nice!

TomAugspurger:
I also deployed things manually ahead of time :) I think the real test will be when we do it for the icesat2 deployment.

salvis2:
I'm working a bit on that now!

PR #626, opened 2020-06-24
Add telemetery

TomAugspurger:
- Adds prometheus and grafana as chart dependencies (affects all hubs)
- Adds default prometheus datasource to grafana (just gcp so far)
- Tell prometheus to scrape a few endpoints for jupyterhub, dask (just
  gcp so far)

TODO (later)

- [ ] make charts, export, and ensure they're imported by default.
- [ ] Hook up grafana to some kind of auth (decide what is public)

cc @jhamman

PR #679, opened 2020-07-25
Enable Metrics and Monitoring on all hubs

salvis2:
Taking config that @TomAugspurger made and putting it into `pangeo-deploy/values.yaml` so that the GCP hub has the same settings, but the AWS gains the same benefits. This is just the changes from #626 and #629 , minus the port-forwarding commands. Need to look more at #648 .

To-Do:
- [x] ~~Port forward~~
- [x] Install `prometheus-operator` as part of ~~the CI for the AWS hub (and ooi?)~~ the `hubploy` local Helm Chart
- [x] Add in DNS records for `ingress-nginx`

Let me know if there are some other things to do to make this happen. Should I try to manually deploy to the AWS hub's staging deployment to test? Or would merging this PR basically be that test and I shouldn't bother? cc @scottyhq . Will probably push a little bit more on this over the weekend.


salvis2:
@TomAugspurger I've looked back at #648 again and was reminded that CI couldn't install `prometheus-operator` for us. Did you have any idea why it wasn't working and if there were possible workarounds?

Edit 1: Ah, it looks like `.circleci/config.yml` doesn't actually issue _any_ `helm upgrade` commands directly. It's all `hubploy deploy`.

Edit 2: Have we considered using a different CircleCI orb (tho right now we are just using a Python image from Docker and installing `requirements.txt`)? I'm seeing that CircleCI has some potential solutions:
- AWS EKS Orb, [Install Helm Chart job](https://circleci.com/orbs/registry/orb/circleci/aws-eks#jobs-install-helm-chart)
- GKE orb has no equivalent command (oops)
- Helm Orb, [Install Helm Chart job](https://circleci.com/orbs/registry/orb/circleci/helm#commands-install-helm-chart), could be generically usable for all clusters?

There appear to be a few options for GitHub Actions, which is something Hubploy supports but we don't use it here. None of these appear to be hugely followed, but the first is the most followed (50 stars vs 3 vs 3).
- [DeliveryBot](https://github.com/deliverybot/helm) with a usage [blog post](https://medium.com/swlh/deploying-to-kubernetes-with-helm-and-github-actions-14825e6df1f2)
- [`kubectl-helm`](https://github.com/marketplace/actions/kubectl-helm)
- [Helm Deploy](https://github.com/marketplace/actions/helm-deploy)

FINAL EDIT: Since we are deploying `prometheus-operator` with the Helm Chart that `hubploy` deploys, this comment is irrelevant to the PR.

salvis2:
From some conversation in https://github.com/pangeo-data/pangeo-cloud-federation/pull/671 with @TomAugspurger , we can just bundle `prometheus-operator` and `ingress-nginx` under the chart that `hubploy` deploys. So we will go with that and set up the config for the AWS hub here as well.

salvis2:
Deployed the above with local `hubploy` commands. I needed to deploy it once to get the `loadBalancerIP` for `ingress-nginx` since that IP isn't set to an existing IP address like it is for the GCP hub.

pangeo-deploy/values.yaml
```diff
@@ -94,6 +117,12 @@ homeDirectories:
 # * ingress-nginx.controller.service.loadBalancerIP: ...
 
 prometheus-operator:
+  alertmanager:
+    enabled: false
+
+  nodeExporter:
```
TomAugspurger, line 123: Do you know what nodeExporter does? This does per-*node* monitoring? But we're mostly interested in per-*pod* statistics?

salvis2: Node exporters should be looking at hardware / OS (if I'm reading this right): https://github.com/prometheus/node_exporter#node-exporter, so yeah, per-node monitoring. We have only really done per-pod stuff up to now I think.

TomAugspurger: Sounds good, thanks.

PR #682, opened 2020-07-27
Bump pangeo helm chart version

TomAugspurger:
This includes bumping dask-gateway to 0.8.0

### DIFF
--- a/pangeo-deploy/Chart.yaml
+++ b/pangeo-deploy/Chart.yaml
@@ -5,7 +5,7 @@ name: pangeo-deploy
 version: 0.1.0
 dependencies:
   - name: pangeo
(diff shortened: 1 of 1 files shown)


jhamman:
@TomAugspurger - do you want to also bump the client-side dask-gateway versions in this PR?

TomAugspurger:
We have dependabot PRs for those, but I'd actually like to test it with 0.7.1 in the client images before merging those. Things should be compatible.

TomAugspurger:
Well, you can create a cluster but workers fail to start because of https://github.com/dask/dask-gateway/pull/282. `dask-gateway-worker` needs to expect the `--dashboard-address` argument.

So we will need dask-gateway>=0.8.0 in worker images when the service has 0.8.0.

PR #684, opened 2020-07-28
store only pangeo-data teams

### LINKED ISSUE #683: Auth0 fails for users with more than 30 github teams
The GitHub API returns paginated responses with 30 results by default. So the rule that we depend on to add all of a user's github teams can fail to capture `pangeo-data/us-west2` or `pangeo-data/us-central1-b-gcp`:
https://github.com/pangeo-data/pangeo-cloud-federation/blob/staging/auth0/2-add-github-teams.js 

An example for iterating over response pages in Python is here:
https://github.com/pangeo-data/pangeo-cloud-federation/blob/d0fc5e00ba455e27c411646e88a0429890b1097d/.github/scripts/update_membership.py#L46-L58 

I think the solution we should adopt those is to only record `pangeo-data` organization teams. Sound good @rabernat @TomAugspurger ?

cc @rsignell-usgs who belongs to an astounding number of github teams ;)

scottyhq:
fixes https://github.com/pangeo-data/pangeo-cloud-federation/issues/683 cc @TomAugspurger 

Still some issue in my attempt here since my Auth0 metadata shows:
```
{
  "roles": [
    "pangeo-data/pangeo-members",
    "pangeo-data/pangeo-admin",
    "pangeo-data/nasa-access",
    null,
    "pangeo-data/pangeo-k8s",
    null,
    "pangeo-data/us-central1-b-gcp",
    "pangeo-data/aws-uswest2"
  ]
}
```
I don't know where the `null` values are coming from...

### DIFF
--- a/auth0/2-add-github-teams.js
+++ b/auth0/2-add-github-teams.js
@@ -7,7 +7,8 @@ function (user, context, callback) {
   })[0];
   var access_token = github.access_token;
   request.get({
-      url: "https://api.github.com/user/teams",
+      baseUrl: "https://api.github.com/",
+      uri: "/user/teams?per_page=100",
       headers: {
         // use token authorization to talk to github API
         "Authorization": "token "+access_token,
@@ -20,7 +21,9 @@ function (user, context, callback) {
         if(data){
           // extract github team names to array
           var github_teams = JSON.parse(data).map(function(team){
+            if(team.organization.login === 'pangeo-data'){
             return team.organization.login + "/" + team.slug;
+            }
           });
           // add teams to the application metadata
           user.app_metadata = user.app_metadata || {};
--- a/auth0/2-add-github-teams.js
+++ b/auth0/2-add-github-teams.js
@@ -1,5 +1,4 @@
 // https://gravitational.com/blog/aws-github-sso/
-// Uses first org in REQUIRED_GITHUB_ORGS application list?
 function (user, context, callback) {
   // access token to talk to github API
   var github = user.identities.filter(function (id){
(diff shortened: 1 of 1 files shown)


TomAugspurger:
> I don't know where the null values are coming from...

Ares these for private organizations / teams?

auth0/2-add-github-teams.js
```diff
@@ -7,7 +7,8 @@ function (user, context, callback) {
   })[0];
   var access_token = github.access_token;
   request.get({
-      url: "https://api.github.com/user/teams",
+      baseUrl: "https://api.github.com/",
+      uri: "/user/teams?per_page=100",
```
TomAugspurger, line 10: This will fix the issue for people on 30-100 teams, but we might still have an issue if people are on more teams than that?

scottyhq: that's correct. this is a quick fix. if someone wants to spend more time figuring out javascript i think all the current rules could be cleaned up and less error prone

TomAugspurger (approved):

scottyhq:
> Ares these for private organizations / teams?

Not sure, rather than get to the bottom of it I went for a different strategy, just filter the list of up to 100 teams for any team containing the string 'pangeo-data' 

Not optimal, but gets the job done for now. My metadata on Auth0 reports the following:
```
{
  "roles": [
    "pangeo-data/pangeo-members",
    "pangeo-data/pangeo-admin",
    "pangeo-data/nasa-access",
    "pangeo-data/pangeo-k8s",
    "pangeo-data/us-central1-b-gcp",
    "pangeo-data/aws-uswest2"
  ]
}
```

Going to merge since this is the current config being used on Auth0

PR #687, opened 2020-07-29
Bump gateway for OOI

TomAugspurger:
Staging is currently deployed with dask-gateway 0.8.0 which unfortunately includes a breaking change for starting up workers (my bad). So we require dask-gateway>=0.8.0 in the worker images. cc @tjcrone.

@scottyhq can you confirm: does anything else in the `deployments/ooi/image/binder/` directory do anything (the environment.yaml file, etc.) since the Docerfile is using `pangeo/pangeo-notebook` rather than `pangeo/base-image`?


TomAugspurger:
Answering my own question: they do seem to be applied still. I must have misunderstood.

scottyhq:
@tjcrone - what are your plans for the ooi image? i remember you mentioning you might build it separately in a different repo?

deployments/ooi/image/binder/Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Note that there must be a tag
-FROM pangeo/pangeo-notebook:afe4262
+FROM pangeo/pangeo-notebook:2020.07.28
```
scottyhq, line 2: this should be `pangeo/base-image:2020.07.28` if you want to manage your own environment.yml etc

TomAugspurger:
I'd like to proceed with this. I think it's the smallest change that will get everyone to dask-gateway 0.8, so that we can deploy prod.

TomAugspurger:
The helm chart refactor that I'm pushing through is going to require dask-gateway>=0.8.0, so I'm going to go ahead with this soon and then do a staging -> prod merge.

PR #706, opened 2020-08-31
add ml image to gcp

### LINKED ISSUE #700: Bring ML image to gcp cluster?
The AWS cluster has a "ML Notebook" image, which points to https://github.com/pangeo-data/pangeo-docker-images/tree/master/ml-notebook. The GCP cluster does not. But many GCP users (e.g. @anirban89) want to do ML from GCP.

Is it as simple as adding this image to the profile list?

rabernat:
closes #700 

cc @anirban89

### DIFF
--- a/deployments/gcp-uscentral1b/config/common.yaml
+++ b/deployments/gcp-uscentral1b/config/common.yaml
@@ -82,8 +82,8 @@ daskhub:
                     'image': 'pangeo/ml-notebook:master',
                     'cpu_limit': 2,
                     'cpu_guarantee': 1,
-                    'mem_limit': '8G',
-                    'mem_guarantee': '4G',
+                    'mem_limit': '16G',
+                    'mem_guarantee': '12G',
                 }
             },
           ]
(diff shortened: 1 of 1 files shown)


deployments/gcp-uscentral1b/config/common.yaml
```diff
@@ -75,6 +75,17 @@ daskhub:
                     'mem_guarantee': '12G',
                 }
             },
+            {
+                'display_name': 'ML Image - Large (12 GB - 16 GB)',
```
TomAugspurger, line 79: I think this doesn't align with the memory limits below.

TomAugspurger:
Thanks!

rabernat:
Ok this works. But I realized that the ml image doesn't have Keras! 🤦

TomAugspurger:
In the meantime, @anirban89, does using `tf.keras` work for you?

rabernat:
Actually keras works. You just have to import it from tensorflow!

```python
from tensorflow import keras
```

PR #709, opened 2020-09-01
Update image target

TomAugspurger:
xref #707


TomAugspurger:
OK, this one fixed it.

PR #720, opened 2020-09-02
Update OOI base image

tjcrone:


tjcrone:
@salvis2 Looks like we are both debugging our deployments at the same time. I suspect this will work fine if we merge our own commits at will here, but please let me know if you would like me to wait until you are finished. Thanks!

salvis2:
I will leave it for a bit. Hack away!

tjcrone:
Okay @salvis2 I'm done merging to this repo for the evening. Thank you!

PR #736, opened 2020-09-17
Up timeout to 1200 and update hubploy

salvis2:
Suggestion from @tjcrone

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -198,7 +198,7 @@ jobs:
           when: always
           no_output_timeout: 1200
           command: |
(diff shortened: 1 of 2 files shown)


tjcrone:
I think you timed out "waiting for the condition". That means that Helm timed out, not the CI output timeout. Add the timeout also to the hubploy deploy command.

tjcrone:
If something else is going on, increasing the timeout may not fix this. I can take a closer look later this afternoon. Sorry for not being able to dig into this right now.

salvis2:
Thank you for looking at all! Moving too fast this morning and missing both instances of `1200`.

PR #739, opened 2020-09-17
staging -> prod

TomAugspurger:


tjcrone:
Not sure I'm understanding this:
```
Error: rendered manifests contain a resource that already exists. Unable to continue with install: existing resource conflict: namespace: , name: ooi-prod-grafana, existing_kind: policy/v1beta1, Kind=PodSecurityPolicy, new_kind: policy/v1beta1, Kind=PodSecurityPolicy
```
There is nothing with this name in an unnamed namespace. That may be the clue.

tjcrone:
I think it may be this: https://github.com/pangeo-data/pangeo-cloud-federation/blob/c7f801b0ebeef0d8f79ab480c32e86862c25f3c6/deployments/ooi/config/prod.yaml#L26
I often delete OOI staging because I don't need it hanging around. Is there a way to make the deployments independent of one another?

TomAugspurger:
> There is nothing with this name in an unnamed namespace. That may be the clue.

That may be an artifact of CircleCI trying to strip sensitive values from the output log. Or we could legitimately be trying to install into a blank namespace, but I think we would have gotten an error earlier.

The grafana stuff is to support dashboards like http://grafana.us-central1-b.gcp.pangeo.io/grafana/d/a_khcwhWk2/dask-2?orgId=1&refresh=5s. I don't see an nginx-ingress, so I suspect that none of it is working. I'll disable metrics entirely on OOI for now, but we can reenable them later if you want.

tjcrone:
Okay that looks neat but I'm okay not having that for now. I see that you disabled it. That works for me. Thanks!

PR #751, opened 2020-09-23
Remove prod monitoring deployment

salvis2:
The last `prod` deployment failed on AWS, surprisingly: https://app.circleci.com/pipelines/github/pangeo-data/pangeo-cloud-federation/1365/workflows/07fdcc01-1446-409c-a963-1a10d9b912f2/jobs/1526. Thanks @JessicaS11 for the heads-up.

Error logs show:

```
Too long with no output (exceeded 20m0s): context deadline exceeded
```

Looking at the cluster on the command line, I see there are `prometheus-operator`-related pods on `prod` now. They have been there for 5d 19h. These should not be there. The `hub` pod is `PENDING` and the core node has 83% cpu requests, so I'm guessing it's just too full.

I have a couple commits in here to remove config that involved `prometheus-operator`. I had assumed we would do a deployment for both `staging` and `prod` but this was not the case. The config I removed was useful _if_ there was going to be a `prod` monitoring deployment but we are holding off on that for now. Removed hosts and grafana admin password. I'm wondering if specifying anything in those "activated" a full `prometheus-operator` deployment.

@TomAugspurger thoughts?


salvis2:
I will delete all the `prometheus-operator` resources in `prod` to get the hub up in the short term. Hopefully this PR fixes the reason that they showed up at all.

salvis2:
Realized @TomAugspurger did this for OOI: https://github.com/pangeo-data/pangeo-cloud-federation/commit/33114ce014f0356076662d8ad7c21868a109f8f9 so I matched it for AWS in the last commit.

TomAugspurger:
Thanks!

PR #774, opened 2020-10-06
Correct build action

salvis2:
So the `build` action is broken. The error right now is roughly:

```
fatal: Invalid symmetric difference expression ee370dc0c3dc2512bad460e0798d6f82d0d58726...HEAD
[...]
subprocess.CalledProcessError: Command '['git', 'diff', '--name-only', 'ee370dc0c3dc2512bad460e0798d6f82d0d58726...HEAD', '--', '/github/workspace/deployments/icesat2/image']' returned non-zero exit status 128.
```

As far as I can tell, this is supposed to mean that `git diff` couldn't find one of the referenced commits. Since we only specify one (the other is `HEAD`), it can't find the only commit hash we provide. I put a step (`Check git log`) into the GitHub Action:

```
git log -n 5
```

to get the last 5 commits. As seen in the Action [here](https://github.com/salvis2/pangeo-cloud-federation/actions/runs/291951994), all I get back is

```
Run git log -n 5
commit 88af23aa956f3f7e6ed501063acbf53487dccc30
Author: Sebastian Alvis <sebialvis@gmail.com>
Date:   Tue Oct 6 11:13:07 2020 -0700

    Check git log to see if it has the commit history
```

So it doesn't have any of the history beyond what I just gave it? I'm setting up this PR to track my work in fixing this. Maybe the PR will find more commits.

### DIFF
--- a/.github/workflows/build-aws-image.yaml
+++ b/.github/workflows/build-aws-image.yaml
@@ -23,6 +23,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
       #- uses: docker://yuvipanda/hubploy:20200826083951674280
       #  name: Unlock git-crypt Secrets
       #  env:
--- a/.github/workflows/build-aws-image.yaml
+++ b/.github/workflows/build-aws-image.yaml
@@ -3,7 +3,6 @@ on:
   push:
     branches:
       - staging
-      - correct-build-action
       #- prod
     paths:
       - 'deployments/icesat2/image/binder/*'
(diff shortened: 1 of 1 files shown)


salvis2:
Okay, I've currently enabled the GitHub Action to see all tags and branches [with this info](https://github.com/marketplace/actions/checkout#fetch-all-history-for-all-tags-and-branches) so that it can always find any commit that it needs to find commit ranges. @scottyhq @TomAugspurger do you think this is fine behavior? I can try to restrict it a bit more if we want but I'd have to dig in a bit.

I need to clean up the PR but it's looking done otherwise.

scottyhq:
> so that it can always find any commit that it needs to find commit ranges. @scottyhq @TomAugspurger do you think this is fine behavior?  

Seems good to me! feel free to merge

PR #784, opened 2020-10-09
Add git-crypt step back into build action

salvis2:
Should fix the problem here: https://github.com/pangeo-data/pangeo-cloud-federation/pull/781#issuecomment-706307046


salvis2:
Recalling this comment: https://github.com/pangeo-data/pangeo-cloud-federation/pull/754#issuecomment-703973995

Forks can't read the `GIT_CRYPT_KEY` secret and thus the error, but we should be fine to merge.

PR #807, opened 2020-10-29
Adjust CPU:Memory ratio for autoprovisioning

TomAugspurger:
The high-memory workers we use have a Memory:CPU ratio of ~6:1. This updates the values in `autoprovisioning.json` to reflect that.

We'll need to manually set this using gcloud commands. It's not set as part of CI.


TomAugspurger:
I've updated this to adjust how we set the worker memory and CPU requests. We now accept a float input and scale the *request* down by some factor. For CPUs we request about 88%, and for memory we request about 95%.

I'll try this out on staging.

PR #813, opened 2020-11-02
Set cluster limits

### LINKED ISSUE #812: High resource usage
@shunzlu right now you appear to be requesting many workers

```
$ kubectl -n prod get pod -l 'hub.jupyter.org/username=shunzlu' | wc -l
8766
```

That's 8766 scheduler / worker pods. At the moment, we don't have a way to limit a single user's usage, so one person can take all the resources of the entire kubernetes cluster, starving other users' resource requests.

Can you check your code to ensure you aren't using too many resources? Perhaps you're using `.adapt()` rather than `.adapt(minimum, maximum)`?

We unfortunately don't have a good way for *you* to know this on your own. There's http://grafana.us-central1-b.gcp.pangeo.io/grafana/d/a_khcwhWk2/dask-2?orgId=1&refresh=5s, but that doesn't scale well and doesn't show per-user usage yet. The proper solution is per-user limits, but that's going to take a while to implement.

TomAugspurger:
This limits the size of the Dask cluster. There's currently no way to
limit the number of Dask clusters per user, so a single user can still
swamp the kubernetes cluster by making multiple dask clusters.

These values were taken from https://github.com/pangeo-data/pangeo-cloud-federation/pull/508, with the memory usage increased to 600GB to reflect the 6:1 memory:CPU ratio of our nodes on GCP.

@tjcrone @scottyhq do these values seem OK to you? If not I can apply them to just GCP.

closes #812

### DIFF
--- a/pangeo-deploy/values.yaml
+++ b/pangeo-deploy/values.yaml
@@ -123,6 +123,10 @@ daskhub:
         idle: |
           # timeout after 30 minutes of inactivity
           c.KubeClusterConfig.idle_timeout = 1800
+        limits: |
+          # per Dask cluster limits.
+          c.ClusterConfig.cluster_max_cores = 100
+          c.ClusterConfig.cluster_max_memory = "600 GB"
 
 homeDirectories:
   nfs:


tjcrone:
It seems reasonable to me under normal circumstances. But sometimes when I am doing a demo I start well more cores than this. Seems like this is something that should be set per deployment, but we can also override so I am not strongly opposed to including this.

scottyhq:
thanks @TomAugspurger ! Definitely good to have this in place. 

For what it's worth on AWS dask worker nodes currently use r5.2xlarge instances which have a Mem:vCPU ration of 8:1 . Also at the dask worker nodegroup level we set a max, which is currently 50 nodes. In effect this is another hard limit on all running clusters (400vCPU, 3.2TB RAM). And finall the number schedulers that can be running simultaneously on scheduler nodegroups (currently tops out at 40 for AWS)... most of the time this more than enough. Workshops have been the only time we've tested the limits.

tjcrone:
```ValueError: 600 GB is not a valid memory specification. Must be an int or a string with suffix K, M, G, T```

tjcrone:
Fixing here: https://github.com/pangeo-data/pangeo-cloud-federation/pull/819

PR #817, opened 2020-11-02
Storage for MLFLow

TomAugspurger:
1. Creates mlflow Google Service Account with write access to
   gs://pangeo-scratch
2. Configures our MLFlow deployments to store logs in a sqlite database.
3. Configures our MLFlow deployments to store artifacts to
   gs://pangeo-scratch/mlflow/

### DIFF
--- a/mlflow/values.yaml
+++ b/mlflow/values.yaml
@@ -1,5 +1,5 @@
 # TODO: set good defaults and override for GCP.
-backend_store_uri: "sqlite:///mlruns.sqlite"
+backend_store_uri: "sqlite:///srv/mlflow/mlflow.sqlite"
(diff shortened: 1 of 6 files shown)


mlflow/values.yaml
```diff
@@ -1,12 +1,15 @@
-backend_store_uri: "file:///mlruns"
-default_artifact_root: ""
+# TODO: set good defaults and override for GCP.
+backend_store_uri: "sqlite:///mlruns.sqlite"
```
TomAugspurger: This has a couple problems. This would break when we deploy to prod, since both would be writing to the same database file.

I should figure out how persistent volumes and PVCs work.

PR #844, opened 2020-11-04
Note workflow

TomAugspurger:


tjcrone:
Thank you for this clarification. I think this is how we have all been working normally, right? Do we want to mention anything about squashing commits (or not squashing)? I work on staging on my fork, and then PR from staging <- staging, so squashing those commits would rewrite history on my fork I think.

TomAugspurger:
I've changed the settings in GitHub to only allow rebase and merge, so we shouldn't need to mention that.

If we care deeply about keeping the history clean we can recommend manually squashing prior to merging so that 1 PR is 1 commit. But it doesn't matter too much to me.

tjcrone:
This is great thanks Tom. I don't have strong opinions about merge strategies either, as long as we don't squash on merge. I usually try pretty hard to build a commit history that tells a story but PRs should be able to have multiple commits if need be.

PR #922, opened 2021-01-28
fix helm lint, circleci helm setup

scottyhq:
should fix current CI failures 

- no longer doing anything with AWS via circleCI (all on github actions)
- use same helm version in circleci and github actions (3.4.1)
- need to change 'stable' repo to https://charts.helm.sh/stable 
- i think the github actions 'helm lint' job needs 'helm dependency update'

cc @TomAugspurger

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -104,10 +104,6 @@ jobs:
             pip install --upgrade -r requirements.txt
 
             curl -sSL https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-226.0.0-linux-x86_64.tar.gz | tar -C venv/ -xzf -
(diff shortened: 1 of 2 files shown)


scottyhq:
`Error: repo "https://kubernetes-charts.storage.googleapis.com" is no longer available; try "https://charts.helm.sh/stable" instead`

CircleCI failed in the helm install step I think due to `helm repo add prometheus-operator https://kubernetes-charts.storage.googleapis.com` but deployment apparently still worked through hubploy...

TomAugspurger:
Mmm strange.

IIRC, I don't think that prometheus is deployed through CD. It's done separately, so we could perhaps delete it from CircleCI if it's causing issues.

PR #957, opened 2021-05-05
bump prod image versions

scottyhq:
it's been a while, not sure if folks want to do this. I find myself always using staging on aws.


rabernat:
Thanks so much for doing this Scott! 🚀

rabernat (approved):

